PR #2, opened 2015-01-18
Fix wrong operator in argument

ufdada:
very unlikely that all conditions are true at the same time...


ufdada:
Sorry, should be fine now. I removed it since i don´t see any value in it.

tonihele:
Is there anything in particular you would like to do? I appreciate the help you have given so far. Currently there is only coarse TO-DO-list on the readme. I could easily produce some smaller issues which would be easy to process. I don't even how to contact you or how willing you are to contribute to this project, please let me know if there is something on your mind.

ufdada:
Well actually i am still messing with the code to get into it ;). I wanted to reconstruct the main menu, but since i realised that even that is "just" an awful complex map, i probably should work on something easier. It would be nice if you could split those big tasks into smaller ones. In most projects i do localisation (in german), so expect something like this to happen :).

PR #24, opened 2015-01-21
import the jme vector

ufdada:
Otherwise it´s a little hard to play arround with it.

### DIFF
--- a/src/toniarts/opendungeonkeeper/tools/convert/kcs/KcsEntry.java
+++ b/src/toniarts/opendungeonkeeper/tools/convert/kcs/KcsEntry.java
@@ -4,7 +4,7 @@
  */
 package toniarts.opendungeonkeeper.tools.convert.kcs;
 
-import javax.vecmath.Vector3f;
+import com.jme3.math.Vector3f;
 
 /**
  * Stores the KCS file entry structure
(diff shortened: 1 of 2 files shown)


tonihele:
You can convert the JAVA vector to JME vector in the actual implementation. But I want to keep the actual DKII data file parsings as pure JAVA. So that they are not tied to any 3rd party engine or library. They are just usable data beans.

The whole convert -package was supposed to be JME free since I thought that everything would be converted to our own formats with would reside in another package hierarchy (like *world.*). And loading would have been native JME. But as I reverse-engineered the actual data and realized there is sooooo much data, this is on hold. Although we are never ever (most likely :) ) switching the engine or anything. But I see a value in keeping them just pure JAVA. If ever there is need for someone to convert them to another language i.e. etc.

But what comes to these KSC, they could be converted to JME motion paths and loaded natively in JME.

There also this, you might not need vectors after all. It it just my retarded idea (I thought I figured them out, but I didn't, most likely...). Best guess at the moment is that the values actually form a bezier (http://en.wikipedia.org/wiki/B%C3%A9zier_curve). At least according to some experts... But it is a guess also....

ufdada:
Oh well, sorry. I thought it was just an importing mistake. Good to know though!

PR #28, opened 2015-02-06
sort offsets (german offsets aren't sorted correctly)

ufdada:
Looks like an really easy change. 
Tested it with all textfiles in german.

### DIFF
--- a/src/toniarts/opendungeonkeeper/tools/convert/str/StrFile.java
+++ b/src/toniarts/opendungeonkeeper/tools/convert/str/StrFile.java
@@ -11,6 +11,7 @@ import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.CharBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


tonihele:
There is a possibility you are right, but I think it is the key that we extract the entries in order. The order which is determined by the offset LIST in the STR file header, not the order determined by the REAL file offsets. That is why I suggested the copy of the array which would be sorted instead the original one (which would be used to process the entries), and binarySearch the next logical offset to get the length. Only way to be sure is to actually compare the problematic German file to the English one. 

Or!!! Better yet, compare with the editor, some ID, see if it matches the line number. I mean is sensible to the object.

The order determines the ID in which they are requested in the game. You can see the dictionary I stubbed, it already converts the assets.

ufdada:
Sorry, i completely misunderstood you. Is the last commit better now?

Maybe i should get sleep :)

tonihele:
Hard to explain these, but you got it! If it works, yes :)

PS. http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#binarySearch%28java.util.List,%20T%29

No need to convert to arrays. You can straight up sort a List. I'm sure about performance-wise. But maybe cleaner code.

tonihele:
Perfect! Thousand thanks!

PR #29, opened 2015-02-07
Extract texture wads

ufdada:
This PR also extracts `EngineTextures.WAD` and `FrontEnd.WAD` to the Textures directory.

It also has a small bugfix for reading wads and extracting them.


tonihele:
What exactly was there in these files? Did the entries override some of the entries from the texture package? They contain probably more recent textures than the texture package. Just curious.

These are probably okay. Ill look them up when I get home.

ufdada:
`FrontEnd` has all gui elements (like loadingscreens, movie icons, level icons, grid for mp etc.) for the frontend included (no duplicates afaik).
`EngineTextures` has a lot of duplicates and as far as i can tell only the map texture (if you press m ingame) is really new.

PR #41, opened 2015-03-14
Typo fix

ufdada:
Fixed a typo (campaign instead of campaing) and converted the if else to switch cases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /nbproject/private/
 /build/
+/assets/Interface/Texts/*.properties
 # property file which holds the configuration
 odk.properties
 *.png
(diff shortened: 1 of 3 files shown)


tonihele:
Looks good. FYI, I'm currently trying to create the graphics options. To get rid of the JME one (branding) and add things like SSAO

ufdada:
Sounds great :). I try to fix some things in the campaign menu (like camera, the arrows, some broken graphics)

tonihele:
If you manage to get the camera right, I'll give you a gold medal :) I thought that I would implement the level camera things on the game state (to test & show) but maybe it is better to wait the proper trigger system implementation. But I think there is a good amount of the camera transitions to get it right on the main menu alone.

The campaign level selection I think is "broken" since there is no logic of campaign advancement. So everything is just laid out there. Especially the A & B & C levels + the "level" probably make them very broken. The animations & amount of "raising" on the levels is just reverse engineered by eye.

Do you use IRC etc? Or even Steam etc. It would make communicating easier if needed.

ufdada:
Here is my steam profile: http://steamcommunity.com/profiles/76561197962486008/

PR #44, opened 2015-03-28
check if conversion is needed (version number has changed) for separate ...

### LINKED ISSUE #43: Expand the asset conversion process to support partial conversions (conversion versioning)
Expand the asset conversion process to support partial conversions (conversion versioning). Conversion process might take awhile and it is often not necessary to convert everything. Current system has only one version number for the conversion that forces whole conversion process to take place.

This system would serve both, the developers and the future end users.

ufdada:
...processes.

Fixes #43

### DIFF
--- a/src/toniarts/openkeeper/Main.java
+++ b/src/toniarts/openkeeper/Main.java
@@ -49,6 +49,7 @@ import toniarts.openkeeper.gui.CursorFactory;
 import toniarts.openkeeper.setup.DKConverter;
 import toniarts.openkeeper.setup.DKFolderSelector;
 import toniarts.openkeeper.setup.IFrameClosingBehavior;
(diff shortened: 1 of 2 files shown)


src/toniarts/openkeeper/Main.java
```diff
@@ -68,7 +68,6 @@
     private final static String DKII_FOLDER_KEY = "Dungeon Keeper II folder";
     private final static String CONVERSION_DONE_KEY = "Conversion done";
     private final static String CONVERSION_VERSION_KEY = "Conversion version";
```
tonihele: Is this const needed after this at all?

ufdada: Ah i knew i missed something ;)

tonihele:
Looks good and simple

PR #47, opened 2015-04-04
make camera clipping distance a little shorter

ufdada:
As the clipping distance is pretty high on jmonkey, objects get pretty fast cut through which makes some ugly graphic issues.

Before:
camera-clip1

After:
camera-clip2

The example was achieved by setting the y component to 0 instead of 0.35f on Line 64.

See also:
http://hub.jmonkeyengine.org/t/how-to-set-the-camera-near-clipping-distance/17661


tonihele:
Grrrrreeeaat!

PR #61, opened 2015-07-21
Implement the credit list

### LINKED ISSUE #55: Add credits to main menu
Add credits to main menu. The real original credits that is. I didn't find them anywhere on the resources. Maybe we just need to copy + paste them from the Internet. And check that all is there. We should still use the translations though (titles), they are in the main dictionary.

Nice & simple task

ufdada:
As the title already says this commit should implement the original credit list from Dungeon Keeper 2.

Todo list:
- [x] Name list
- [x] Titles (like Producer) should be taken from the translations
- [x] Infinite autoscroll
- [x] Names should only scroll in the middle part and not the whole screen
- [x] Ok-Button should work again

Fixes #55

### DIFF
--- a/assets/Interface/MainMenu.xml
+++ b/assets/Interface/MainMenu.xml
@@ -472,720 +472,720 @@
     <!-- +++++++++++++++++++++++++++++++++++++++ -->
     <screen id="credits" controller="toniarts.openkeeper.game.state.MainMenuState">
         <layer id="layer" childLayout="vertical">
-            <panel id="panel1" height="27.9%" width="100%" childLayout="vertical">
+            <panel id="panel1" height="12%" width="100%" childLayout="vertical">
                 <panel childLayout="center" align="center" valign="center">
                     <text id="creditsTitle" style="menuTitleText" text="${menu.372}" />
                 </panel>
             </panel>
+            <panel childClip="true" align="center" valign="center" childLayout="vertical" height="70%" visibleToMouse="false">
+                <panel id="creditList" width="60%" name="creditList" align="center" valign="center" childLayout="vertical">
+                    <panel childLayout="vertical" width="100%" height="150%" />
+                    <panel childLayout="horizontal" width="100%">
+                        <text style="menuText" text="Colin Robinson" textHAlign="left" width="50%" />
+                        <text style="menuText" text="${menu.2733}" textHAlign="right" width="50%"/>
+                    </panel>
+                    <panel childLayout="horizontal" width="100%">
+                        <text style="menuText" text="Nick Goldsworthy" textHAlign="left" width="50%" />
+                        <text style="menuText" text="${menu.377}" textHAlign="right" width="50%"/>
+                    </panel>
+                    <panel childLayout="horizontal" width="100%">
+                        <text style="menuText" text="Pete Blow" textHAlign="left" width="50%" />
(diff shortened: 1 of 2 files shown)


tonihele:
Nice. I know how you can achieve the middle part. You can just wrap the creditList inside a panel. Like:
<panel  backgroundColor="#ff0f" childClip="true" height="60%" width="100%" align="center" valign="center" childLayout="vertical" visibleToMouse="false">

childClip="true" might be the key there.

ufdada:
That did the job :+1:
Now i just have to find out why the ok button doesn't work

tonihele:
I commented out the effect and it worked. So my guess is that the button needs to have it's own method that first stops the effect, then calls the doTransition.

ufdada:
Hmpf i can't get it to work. The only thing that makes it at least a little bit better was to move it to another layer (the hover animation did work then), but the not working onclick is still a mystery...

It even seems that normal buttons don't play the sound as long as there is a click attribute. Nifty is really weird :/

tonihele:
You can try the renderOrder on the button panel, if the clipping still doesn't prevent the panel. Or then it is just that the effects block all. I've also noticed the sound playing. Sometimes it seems to work and sometimes not.

Nifty is a bit pain sometimes. But I'd stick with it for now at least. It is still the default in JME 3.1. From my experience they all suck more or less and eventually you need to heavily extend it anyway.

tonihele:
Hmm, onStartScreen seems that is a blocking effect. onActive works and it is not.

tonihele:
And I also noticed that https://github.com/tonihele/OpenKeeper/blob/master/src/toniarts/openkeeper/cinematics/Cinematic.java#L227 breaks everything (everything == trying to do the same transition again). So don't mind it, you'll probably notice it while testing. I'll fix it today.

ufdada:
> Hmm, onStartScreen seems that is a blocking effect. onActive works also and it is not blocking.

Thanks, i will try as soon as i am home again.

Is there a good nifty documentation i could rely on? Everything i find is somehow broken or incomplete.

tonihele:
I have a huge manual for 1.3.2. It's quite good, not by any means complete. Like this, it was just trial and error. Maybe it's logical, maybe not :)

The manual is here:
http://sourceforge.net/projects/nifty-gui/files/nifty-gui/1.3.2/nifty-gui-the-manual-1.3.2.pdf/download

It's hard to find any good examples anywhere by Googling, Sometimes in JME forums. From the alternatives JavaFX might be the next best documented & examples available. But it is nowhere near yet for usage, at least with JME.

ufdada:
The documentation isn't too bad, at least it's better what i found :). onActive did the job quite nicely, i am happy with it right now, what do you think?

tonihele:
Yeah looks good. Just one problem, if you try to go there the second time -> infinite loop.

tonihele:
It's calling the restart credits like mad. Maybe reset / destroy the effect in onEndScreen(). You maybe have a logical loop there :) If nothing helps, then just dynamically construct the scrolling part with the effects every time, or just the effect.

ufdada:
:/ i see, it seems it calls the endeffect even on switching screens (and the effect still running). Didn't test it before because the cinematics could be called just one time. Should be ok now.

tonihele:
It works, I'll merge? Ok?

ufdada:
Sure :)

PR #63, opened 2015-07-25
Implement movie menu

### LINKED ISSUE #57: Add the view movies to the main menu
Add the view movies (extras) to the main menu. It would be nice if part of the logic would also be there (can't watch movies the player hasn't unlocked yet). Of course this logic is implemented just GUI wise here. Since there is no progression nor play yet.

ufdada:
As the title already says, this is the implementation of the movies menu.

What's done so far:
- dynamic generation of the movie buttons
- stub for checking if the movie is unlocked
- showing a message if the cutscene is still locked
- linked the trailer in the extras menu to play the dk3-trailer

Fixes #57

### DIFF
--- a/src/toniarts/openkeeper/game/state/MainMenuState.java
+++ b/src/toniarts/openkeeper/game/state/MainMenuState.java
@@ -109,7 +109,13 @@ public class MainMenuState extends AbstractAppState implements ScreenController 
     private final MouseEventListener mouseListener = new MouseEventListener(this);
     private Vector3f startLocation;
     private static final Logger logger = Logger.getLogger(MainMenuState.class.getName());
-    HashMap<String, String[]> cutscenes = new HashMap<>();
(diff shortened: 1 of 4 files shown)


src/toniarts/openkeeper/game/state/MainMenuState.java
```diff
@@ -103,6 +108,8 @@
     private KwdFile kwdFile;
     private final MouseEventListener mouseListener = new MouseEventListener(this);
     private Vector3f startLocation;
+    private static final Logger logger = Logger.getLogger(MainMenuState.class.getName());
+    HashMap<String, String[]> cutscenes = new HashMap<>();
```
tonihele: Looks great! Only thing I can nitpick is that this could be a static final and instanced in static {} instead of initialize() :)

tonihele:
Very nice. OK from your behalf?

ufdada:
i am ready for merge

PR #67, opened 2015-07-29
Gui improvements

ufdada:
At first i tried to unify the position of the cancel and ok button, since they always keep the same position in dk2. The campaign level selection should now look a little bit better since the images aren't aligned to left anymore.

image


tonihele:
Niccccceeeeee

PR #69, opened 2015-07-31
Implement MPD and Secret level selection to main menu

ufdada:
- MPD Level names are translated now
- Level, Secret level and MPD are using the same campaign screen
- Small GUI Improvements

### DIFF
--- a/assets/Interface/MainMenu.xml
+++ b/assets/Interface/MainMenu.xml
@@ -131,7 +131,7 @@
                 </panel>
             </panel>
             <panel childLayout="horizontal" align="right" valign="bottom">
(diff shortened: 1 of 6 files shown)


ufdada:
I am ready to merge now

tonihele:
Great work dude!

ufdada:
Thanks man :). I guess the last change was for the failing mpd levels? Wonder if they implement this special room with id 35 directly to the maploader...

tonihele:
Some normal levels fail as well, at this point it is not really surprising :) But the now added logging and not completely failing helps in finding the bugs & not implemented parts.

PR #75, opened 2015-08-02
construct hero stone bridge also as a stone bridge

ufdada:
The hero stone needs also to be build like the stone bridge (used in Maps like Alcatraz)


ufdada:
Seems like they had this issue in the original as well :wink: 
Hero stone bridge

tonihele:
Amateurs :) Bridge arches aren't supposed to be build like that, even I know it and I don't even have the education needed to build bridges.

Our reimplementation in my opinion should fix clear bugs.

PR #76, opened 2015-08-02
Basic implementation of lava & water, resolves #73

### LINKED ISSUE #73: Implement water & lava
Implement water & lava. Yes.

tonihele:


ufdada:
Cool stuff :)

tonihele:
Thanks. The lava is super boring now though. In DK II it has waves. I think it maybe more "realistic" to just make it a slow "river" :) Although it doesn't build up to the end as would lava then probably do... Well, anyway, merged in a bit hurry, since I suspect I'll be out for a week or so. So to let everyone play & improve it.

PR #91, opened 2015-10-11
Feature 15

meltzow:
Imps are now loaded and displayed

### DIFF
--- a/src/toniarts/openkeeper/game/state/GameState.java
+++ b/src/toniarts/openkeeper/game/state/GameState.java
@@ -20,18 +20,23 @@ import com.jme3.app.Application;
 import com.jme3.app.state.AbstractAppState;
 import com.jme3.app.state.AppStateManager;
 import com.jme3.asset.AssetManager;
+import com.jme3.bullet.BulletAppState;
 import com.jme3.input.InputManager;
 import com.jme3.renderer.ViewPort;
 import com.jme3.scene.Node;
+
 import de.lessvoid.nifty.screen.Screen;
+
 import java.io.File;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import toniarts.openkeeper.Main;
 import toniarts.openkeeper.game.state.loading.SingleBarLoadingState;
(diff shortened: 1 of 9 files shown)


src/toniarts/openkeeper/game/state/GameState.java
```diff
@@ -103,6 +109,11 @@ protected void updateProgress(int progress, int max) {
                             setProgress(0.25f + ((float) progress / max * 0.75f));
                         }
                     }.load(assetManager, kwdFile));
+                    
+                    bulletAppState = new BulletAppState();
+                    this.stateManager.attach(bulletAppState);    
+                    rootNode.attachChild(new ThingLoader().load(bulletAppState, assetManager, kwdFile));
```
tonihele: I'm afraid this crashes the system. OnLoad() is called from another thread. See how the "worldNode" is loaded separately, only attached to the root node later. We still have the scenegraph to be figured out, but for now you could just attach the ThingLoader stuff to the "worldNode" and it'll work.

src/toniarts/openkeeper/world/MapLoader.java
```diff
@@ -85,18 +88,18 @@ public Spatial load(AssetManager assetManager, KwdFile object) {
 
         // Go through the map
         int tilesCount = object.getWidth() * object.getHeight();
-        Map[][] tiles = object.getTiles();
-        for (int x = 0; x < object.getWidth(); x++) {
-            for (int y = 0; y < object.getHeight(); y++) {
+        Tile[][] tiles = object.getTiles();
+        for (int width = 0; width < object.getWidth(); width++) {
+            for (int height = 0; height < object.getHeight(); height++) {
```
tonihele: And while at it, I don't understand the need to change the variable names here. As this really IS a coordinate in 2D system, and x & y are totally valid in that sense. "height" and "width" really aren't as much valid. But I'll let you decide.

src/toniarts/openkeeper/game/state/GameState.java
meltzow: fixed now

PR #102, opened 2015-11-07
Gui update

ArchDemons:
I tested at resolutions 640x480 and 800x600 - it looks great


ufdada:
Looks good to me :+1:

Issue #98, opened 2015-11-01
Implement creature picking up


tonihele:
Implement creature picking up. We should be able to pick creatures up and drop them. This should be visually shown. No need to implement this to the creature cards yet, since they are not existing. Just the mechanics and visuals.

The picking up is a stack. Last in first out?

tonihele:
Implemented as an interface, no objects still fully support this. Evictions / sacrificing is not implemented, but UI wise there is an enum prepared for this. The creatures look funny because they are in the GUI node which looses any depth. No icons either when picked stuff up. Nor actual falling down or being stunned, mainly since these require more work (like how do we implement physics...).

Issue #97, opened 2015-11-01
Implement creature slapping


tonihele:
Implement creature slapping. SLAP!

Can slap. Slapping should result in visual violence. The creature needs to know it has been slapped. No mood etc changes required. Nor blood decals. 

Probably waiting for #16 to be really usable.

Birdasaur:
This is the single most important feature in the entire game. All our hopes rest on you.

tonihele:
Initial version done

ufdada:
really looking forward to it :+1:

tonihele:
There is no direction yet and they kinda tend to just continue their business afterwards :) But getting there...

ufdada:
As long as the sound is playing i would be happy at first glance :)

tonihele:
Implemented the Keeper sound but no cursor (as it is not so simple yet, need to create timer for changing cursor). No direction nor decals nor creature sound (as I don't want to hard code this). Wrong animations. But you can actually kill your creature, whether that is a good thing then...

PR #103, opened 2015-11-07
Mouse cursors

ufdada:
You can't slap your creatures without the correct mouse cursor! ;).

What's done:
* refactored the current Cursorfactory (since the existing cursors shared quite a lot code)
* added the rest of the cursors
* fixed the width issue (just needed to be a multiple of 16)
* fixed the empty line issue (same problem with multiple of 16)

What's not done:
* `NO_SPELL_POSSESS` and `SPELL_POSSESS` have some issues because the file format, so i left them open. If we can't fix them we just build our own, as they are pretty easy.
* The hotspot point may not be correct, they were just a random guess

Should help #98, #97


tonihele:
Grrrrreeeat! Ready?

ufdada:
Yes, should be ok now.

PR #119, opened 2015-11-21
Kwd

ArchDemons:
I would have thought about some of the static definition lists (creatures, objects, effect, spells ...).
And lazy loading (Loading at the time when the information is requested)


ufdada:
Looks good to me, anything to complain? @tonihele

tonihele:
Lot to take in, I'll try to review it a bit better later this week. Probably ok. Thanks for reviewing it. It is important.

PR #121, opened 2015-12-03
Feature 110

ArchDemons:
I played with possession is very rare, could have missed something

### DIFF
--- a/assets/Interface/GameHUD.xml
+++ b/assets/Interface/GameHUD.xml
@@ -280,11 +280,8 @@
         <layer childLayout="vertical">
             <!--The upper panel-->
             <panel height="72px" width="100%" align="left" valign="top" childLayout="center">
-
-                <panel width="100%" childLayout="horizontal">
-                    <panel height="65%" width="*" valign="center" backgroundImage="Textures/GUI/Windows/Bar-00.png" imageMode="repeat:0,0,128,64" />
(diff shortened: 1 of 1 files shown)


tonihele:
The code looks ok, didn't test but I trust it is ok. We can tweak the possession GUI later. I've seen few videos, it is more complex than I remembered :) I didn't also use it much. Apparently you can take the thief for a gig and everything :)

tonihele:
Oh, the issue is by @ufdada. I let him do the honors and see if the issue is satisfied :)

ufdada:
Seems fine, the only thing i could complain are the changes made to the map. Even it is closer to the original scale, it now looks awfully pixelated. What do you think @tonihele ?

image

ArchDemons:
Map was originally drawn in small size in difference from other elements

ufdada:
That's correct, but i think it was somehow intended in the first place to look better.

tonihele:
I just back in the day started to create the gui as scalable. Using
percentages. But this is no longer the case. The map was the last
stronghold of these past days. So in that way the change is welcomed. It is
more original and inline with the rest.

If the picture is more pixelated than it should it maybe of the scaling. By
default Nifty does it quick.. and dirty. There is a flag for this. I used
it somewhere... but it should be set globally. And I dont know it the
latter is possible. Are they scaled propotional to the other elements?

ufdada:
Yes, the scaling is pretty much perfect (it aligns very well with the tab buttons).

tonihele:
filter="true" in XML and when loading images straight with Nifty it is a parameter of the same name. Left one is loaded with filter=true
filter

PR #126, opened 2015-12-11
Feature 109

ArchDemons:
known bugs:
* animation of rooms, spells, traps and doors buttons not work after leaving the possession
* dublicated contorl triggers (player state and posssesion state). I fixed it, but ugly. Need refactor.

### DIFF
--- a/assets/Interface/GameHUD.xml
+++ b/assets/Interface/GameHUD.xml
@@ -277,6 +277,10 @@
     </screen>
 
     <screen id="possession" controller="toniarts.openkeeper.game.state.PlayerState">
+        <!-- creature filter -->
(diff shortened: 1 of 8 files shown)


tonihele:
Looks very good!
 - After exiting the possession, the selected -effect on rooms, spells etc. breaks. It seems that private void updateSelectedItem(InteractionState state, int id) uses screen variable. And it it still possession, not HUD, so we need to use getCurrentScreen? I'm not sure what causes this, or is it just the way Nifty works...
 - States maybe should use the enabled state to manage the key listeners, not to overlap each other

I can accept this as such or do you wish to fix the mentioned points?

PR #157, opened 2016-02-28
Feature 137

ArchDemons:
There are two different ways to flash tiles (in each commit). What way is better ?
* Control on ActionPoint
* Control on WorldState

### DIFF
--- a/src/toniarts/openkeeper/game/action/ActionPoint.java
+++ b/src/toniarts/openkeeper/game/action/ActionPoint.java
@@ -31,7 +31,7 @@ import toniarts.openkeeper.tools.convert.map.Thing;
 public class ActionPoint extends Container {
 
     private final int id;
-    private final ActionPointState parent;
+    private ActionPointState parent;
     private final int triggerId;
     private final Point start;
     private final Point end;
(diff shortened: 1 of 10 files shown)


ArchDemons:
do not merge. Vote first. I`ll add something else (maybe)

ufdada:
@ArchDemons Is this PR still relevant?

If the WorldState would be in control of the flashing, wouldn't it show up for all players? I don't think that would be a good idea for multiplayer (if it is used there...)

tonihele:
It could be used there. It is just the map design. Every Keeper needs their own map anyway. Now we are just using the master map. Flashing should be eventually then moved the player map. And tagging etc.

We just haven't planned this much yet.

ArchDemons:
Still Relevant. I'm on vacation and I finished this when I get back.

ufdada:
No problem, i was just wondering because it was already 5 months old.

Enjoy your vacation! :)

ArchDemons:
rebased :)

tonihele:
Look good to me. Didn't test it though.

ufdada:
I'll test it.

ufdada:
Okay checked it, solid blocks are getting highlighted as they should. But there is still no highlighting on path areas or rooms (for example the room building tutorial in level 1).

tonihele:
Thanks, that is probably okay. I think this is more a refactor & stabilization than a feature addition.

ArchDemons:
> Okay checked it, solid blocks are getting highlighted as they should. But there is still no highlighting on path areas or rooms (for example the room building tutorial in level 1).

Fixed. I forgot about highlighting tiles for building rooms

ufdada:
Looks great now! :)

PR #158, opened 2016-02-28
Effect Loader (initial state)

ArchDemons:
effect loader with test on dungeon heart
added dungeon heart plug
added dungeon heart destroyed model
added dungeon heart decayed model


tonihele:
Cool :+1:

Issue #147, opened 2016-02-20
Implement multi threaded framework for game logic


tonihele:
Implement multi threaded framework for game logic. We currently run pretty much in render queue. This isn't very nice in the long run. The faster we change this, the smaller the problems will be. We need some kind of pauseable scheduler thread pool that runs some of the app states in a single separate thread (executes their render method). Scheduled to run in game tick time (the level has this data, usually 4 == 250ms).

Also outside this we need separate threads for path finding for instance. LIBGDX-AI has some schedulers for this. They might be suitable for us or not.

tonihele:
We have now 2 threads idling. Executing movement (every 20ms) and game ticks, the logic, at specified tick rate (250ms by default). Currently very light but movement will get heavier with collision detection. And game ticks as the AI progresses, now it does do a lot of path finds, but with these map sizes... Not all the stuff have been moved to the GameLogic thread that needs to be there.

Now the movement is better and no longer while in break point etc. the creatures warp millions of miles from the map. Since it is now constant. Also the pausing is a bit better, but there is a catch (I'll make an issue of it)...

Playing some OpenKeeper:
![openkeeper_benchmark](https://cloud.githubusercontent.com/assets/8344766/16425703/800bd1dc-3d6e-11e6-9cc5-6b978741ff90.png)

tonihele:
And of course the logic between the AI & animations is a bit harder to manage now. We'll improve it as we go...

ufdada:
Wow. Were does the spikes of the used ram come from? Is this the garbage collection?

tonihele:
Yes, for me at least this is very normal Java behavior. And pretty nice CPU usage. But for games, I don't know how much memory should be eating / second. Around 350MB of used heap it kicks the GC in which I can't at least see in the game at all, which is a good thing. The GC is the default one in JAVA 8.

ufdada:
I don't think we will get close to the 32 MiB that dk2 required back in the days ;). This already looks decent to me, since it shouldn't be a problem on a modern computer.

tonihele:
We require at least 132MB and then the GC would give a bit headache :) This is still without any intentional optimization so to speak.

Hey, BTW, can you check that Travis? He has been awfully quiet...

ufdada:
Looks like the settings of travis weren't set correctly. Now it should build on pushes and pull request again.

And sorry for the build error that you may have received while i tried the building process!

tonihele:
It's cool! Thanks!

PR #171, opened 2016-06-21
Feature 148

tonihele:


ufdada:
Fantastic work done here!

tonihele:
Thanks! A bit slow since a lot of checks are now done in every frame. But with this https://github.com/tonihele/OpenKeeper/issues/147 the checks should be less frequent and some HashMap accesses could easily be reworked.

PR #178, opened 2016-07-03
Implement cheats

### LINKED ISSUE #168: Implement cheats
Implement cheats. The cheat codes come in handy in testing. No need to create any temp stuff, just the cheats straight to fill in the coffers etc.

I think implementation wise this is really simple. CheatState, listen to the invoke, play the sound, listen for the cheat code and BAM! Constructor just needs to take in the player to whom the cheats are applied to.

ufdada:
This fixes #168

### DIFF
--- a/src/toniarts/openkeeper/game/player/PlayerGoldControl.java
+++ b/src/toniarts/openkeeper/game/player/PlayerGoldControl.java
@@ -63,8 +63,9 @@ public class PlayerGoldControl extends AbstractControl {
         if (value < 0) {
             return;
         }
-        gold += value;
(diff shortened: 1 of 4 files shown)


tonihele:
Nice work!

ufdada:
Time will tell if the abstract function was a good way to go (instead of integrating it directly to the cheatstate) but it should be fairly easy to refactor it if it's not the case.

tonihele:
Most importantly it is implemented now :) We could advance our economy now. Pretty straight forward to bill the player on building. And even IT IS PAYDAY!

PR #179, opened 2016-07-05
Feature 144

tonihele:


ufdada:
Great work, i never knew that the job icons for digging, claiming and reinforcing were actually used ingame :)

tonihele:
I hope they were :D I tried to reverse engineer from YouTube video, I'm on my Linux laptop right now. Trying to avoid Wine etc.

ufdada:
Even if not i would leave them in anyway.

tonihele:
+1

They really help now to see what the little things are up to at least.

PR #182, opened 2016-07-10
Feature 160

ArchDemons:
some base network stuff


ArchDemons:
Sorry it took so long. I change job and do not know how much time can be given to the project in the future. I'll make small changes at least

tonihele:
Thanks, I'll try this in a minute. Good luck with your new job!

ArchDemons:
I joined to the java programmers in my new work :)

tonihele:
Oh nice :) You have a good reference here :)

tonihele:
Nice work! Creating a server seems to work only one time though. We'll get to that later. At least now we have some base.

ArchDemons:
I could not solve this problem. There's really them even more ... I did not understand, whether it is my curve hands, or errors in game engine

PR #183, opened 2016-07-11
Dungeon keeper folder & settings

ufdada:
I tried to squash the duplicated code and refactored it into two util-classes.

Also the dungeon keeper folder is now just an optional parameter.


tonihele:
Nice, with Netbeans & Gradle especially the parameters were a bit pain... Didn't test but looks ok.

ufdada:
Yeah that was bugging me a lot, took me a while to get used to it (gradle with parameters). Maybe i write a small introduction in our wiki, since it isn't really self explaining...

tonihele:
Wouldn't hurt at all to have some writing. The parameter thing is custom (well, I copied it from somewhere, but still not a standard solution), so the only way you figure it out is to read the build file :(

PR #188, opened 2016-07-12
Map Selector

ArchDemons:
some isolation of map selection


ufdada:
Looks good to me :+1:

PR #189, opened 2016-07-15
Misc commits

ufdada:
I know, usually we want just one feature per pull request, but these commits are so small, it shouldn't be a big problem to review them.

I also noticed that the triggers and timers are limited in [GameState](https://github.com/tonihele/OpenKeeper/blob/master/src/toniarts/openkeeper/game/state/GameState.java#L150-L157).
Is there a reason for that? Because some maps (like mpd or secret) try to get values exceeding these values and getting a NPE.


tonihele:
I guess they could be just limitless, no need to initialize them beforehand. I don't know if there is some flag somewhere on the triggers that would tell the amount if there is a need to init them, like the timers. @ArchDemons is our trigger expert.

PR #190, opened 2016-07-16
Feature 170

ArchDemons:
added camera rotation control around action point
PlayerTriggerControl and PlayerManaControl moved to Keeper
fixed TriggerControl action and activation
removed debug interactions (claiming and digging)

some known bugs: 
crash game in level 1 (access from other thread). I think this is some trigger action

### DIFF
--- a/src/toniarts/openkeeper/game/control/Control.java
+++ b/src/toniarts/openkeeper/game/control/Control.java
@@ -32,6 +32,9 @@ public abstract class Control implements IControl {
     protected boolean enabled = true;
     protected IContainer parent;
 
+    /**
+     * empty serialization constructor
+     */
     public Control() {
     }
 
@@ -57,6 +60,7 @@ public abstract class Control implements IControl {
 
     /**
      * To be implemented in subclass.
+     * @param tpf time per frame
      */
     protected abstract void updateControl(float tpf);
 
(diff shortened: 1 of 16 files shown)


ufdada:
Why did you remove the debug interactions?
I still think they are quite useful for testing things (like triggers) without having to play the whole map.

Maybe we should hide them better later (like using a special cheat code or console command).

ArchDemons:
removed only claiming and reinforcing tiles. This options already tested and we have imps with AI :)

ufdada:
That's right, but if you want to claim or dig something which isn't directly reachable for your imps you may have to wait several minutes or it isn't even possible (like if there are enemies around and your imps are running away the whole time).

tonihele:
Looks good, but @ufdada has a point about the debug features. Maybe we should keep those at least for awhile longer.

tonihele:
> crash game in level 1 (access from other thread). I think this is some trigger action

Currently creating a party can cause this. Not much else I guess. The spawn creature should always create the creature to the internal registry and queue the graphical presentation for later. The movement too is detached from the visual presentation. A known issue I'll try to sort soonish.

Physics introduction may complicate this further because it is synced with the graphics as far as I know.

ArchDemons:
all done

tonihele:
Great!

PR #192, opened 2016-07-21
Fixed NPE on dkIIFolder variable by using passed in path param

belgianguy:
The wrong parameter was passed in, which caused an NullPointerExcepion as dkIIFolder was not yet initialized (and path is passed in from the user's file select window)


ufdada:
Shame on me, that's a regression of my refactoring.

Maybe the variable dkIIFolder also has to be set, like
```java
dkIIFolder = PathUtils.fixFilePath(path);
PathUtils.setDKIIFolder(dkIIFolder);
```
Could you check that?

belgianguy:
Hi ufdada, I've checked the code and now in Main it'll always use the convenience method, the other commits are more aestethical (and I feel I should have put them in their own branch).

The second commit streamlines the getting of the DKII Folder in the Main function, having a convenience method is sufficient here, as many other classes rely on that function to get the DKII path. But I do see that many other classes also each define a dkIIFolder variable, is this a performance related instantiation maybe?

The third commit has to do with refactoring string constants that are reused into their own little class, as I'm on Linux (Ubuntu) a folder named "Sfx" and "sfx" are two different things, and when it was converting music, it crashed with an NPE on the Sfx folder...

ufdada:
Well most of this code comes from the early days, where the conversion process wasn't working completely and was for easier testing (by executing the conversion separately). I think Toni wrote most of it on windows, so case sensitivity wasn't a big deal back then ;). The assets have the very same Problem (even in the wad-files the cases are different/wrong) which is solved by getCanonicalAssetKey of ConversionUtils, maybe we should implement it here also.

I wouldn't have made an extra class for the folder constants (DkIIFolderConstants) and just put them into the PathUtils (because they are mostly used in combination of it).

BTW: Your DkIIFolderConstants class also lacks the license, would you mind to add it?

tonihele:
> I wouldn't have made an extra class for the folder constants (DkIIFolderConstants) and just put them into the PathUtils (because they are mostly used in combination of it).

Yeah, after these I'd say it is good to go :) No need for an extra class, it just distributes the same stuff too much.

Using those constants is good, but like @ufdada said, we always need to use the getCanonicalAssetKey since different versions have different casings on the names. My Kubuntu + DK II GOG English version fails to work after this patch. I've changed only some of the AssetConversions stuff to use the said method (based on reports from people with different versions). They should all be changed to fix this once and for all. But this is already fixes a regression & makes the code a bit smarter.

belgianguy:
Okay, that's understandable, I'll put the constants in the PathUtils class then.

ufdada:
Are you ready for merge?

BTW: you can get to the latest base version with
```
git fetch --all -p
git rebase tonihele/master
git push
```
So you don't have to deal with the merging

belgianguy:
I am ready. Feel free to suggest amendments if something is not yet clear/correct. :)

tonihele:
Essentially it is great work. There is just much noise due to merging? If it is not too much trouble and it is possible it would be great to only have the actual changes. Now it touches the map selection etc. Maybe it is possible with rebase etc. I don't know. @ufdada can maybe help with that.

belgianguy:
thanks tonihele, I'll ask @ufdada to see what I can do to make it a more clear cut PR that doesn't touch anything outside of its scope. I just pushed my last changes to my github branch, but I'll investigate how I can remove the changes like the map selection from this PR.

belgianguy:
git rebase upstream/master did work, but it too generates a merge conflict on MainMenuState.java

ufdada:
You could try to kick them out with interactive rebasing.

`git rebase -i`

Doing something like that for the first time is always quite a pain...

ufdada:
i cherry picked your changes and rebased them

belgianguy:
Thanks @ufdada, while I was trying to clean up, I'm afraid I got a bit lost in the git woods and made more mess than I cleaned up, still investigating what wrong turn I made.

tonihele:
Thanks both of you

PR #207, opened 2016-07-30
refactored setting op LoopMode CYCLE and LOOP

belgianguy:
Now both ModelViewer and CreatureLoader use the same method to determine if the LoopMode is CYCLE or LOOP


tonihele:
This is good, but a conflict must be solved. There is one part of @ufdada settings/dk folder changes re-applied here, it might be that?

belgianguy:
The conflict was indeed with the DKII folder code, it has been resolved.

ufdada:
you actually removed your bugfix for the Animation NPE

belgianguy:
I looked at the merge window and thought that the args[0] check caused it. And I thought that I should nit merge the spatial children check, as it was already in another PR.

ufdada:
if you rebase (or merge from master) you actually have to take all changes into consideration.

The best way to check if you didn't do anything wrong is the "File changes" tab, it shows the differences to the branch you want to merge.

PR #210, opened 2016-07-31
Fixed NPE in ModelViewer when viewing a Map

belgianguy:
The EffectManagerState used another KWD variable (the class variable) rather than the local variable.

This caused NullPointerExceptions when loading a Map.


ufdada:
That was the cause for the dungeon heart not showing up...

Thanks!

PR #211, opened 2016-07-31
check if spat has children before we try to get them

belgianguy:
This is an issue in ModelViewer when selecting Terrain in the top dropdown, and a room type in the list dropdown (e.g. Treasury or Lair), these don't have children, so a NPE was produced when the getChild(0) call is done on these.


ufdada:
This change shouldn't hurt, but the whole terrain thing needs to be reworked anyway.

It still needs:
- support for fixed / dynamic rooms
- support for quad-tiles
- support for alternative materials
- support for different maps (currently it just loads the terrain of alcatraz)

PR #212, opened 2016-07-31
Effects in modelviewer

belgianguy:
Added effects to ModelViewer, some are immedeatly visible, but they don't loop, but others don't appear yet (we handle only Image type atm). The location is hardcoded for now, just zoom out until the whole map is visible and pick one, "Mercenary Enter Effect" is a nice big one that is easily spotted.As is "inferno test launch"

### DIFF
--- a/src/toniarts/openkeeper/tools/modelviewer/ModelViewer.java
+++ b/src/toniarts/openkeeper/tools/modelviewer/ModelViewer.java
@@ -56,10 +56,7 @@ import de.lessvoid.nifty.spi.render.RenderFont;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


tonihele:
Add the EffectManager as a state in model viewer. That will make the
effects loop. As it will call the update method. Some manual cleaning for
the effects would be then required however...

ufdada:
A little bit OT, but i see you got the commit from #211 in here, too. Just a little advice from my side: before making another branch switch back to your master branch, update and rebase it and then create another branch.

in console:
```bash
git checkout master             # switch to master branch
git fetch --all -p              # update all remote branches and prune obsolete branches
git rebase <remote-branch>      # (optional) rebase your master branch with another one, like tonihele/master
git branch new_branch           # create new branch
git checkout new_branch         # switch to new branch
```

belgianguy:
Ah, my bad, I thought I had made a clean branch, I'll try to fix it.

ufdada:
You can also just make a new branch like mentioned above and cherry-pick the change you want to add

`git cherry-pick af8f27af23cc84b4dba2c143159f25fa478b731d`

belgianguy:
PR 213 supersedes this one.

PR #213, opened 2016-08-02
Feature 166

belgianguy:


ufdada:
Okay at first sight i have two small improvement requests:
- when switching to something else (like maps) the effect should be removed
- when switching back from maps, the initial marble floor should be seen and the map should be removed

optional
- removing the effect immeditately if another (or the same) has been selected in the list box

belgianguy:
Okay, thanks for the feedback!

ufdada:
No problem, great commit so far!

belgianguy:
The optional objective should already be in place, as I clear all effects as soon as another one is loaded, did you face any issues with that?

What I am adding:
To clear the effects, I've made the clearActiveEffects method public, and I clear them whenever a new upper listbox change is made.

ufdada:
By removing them from the arraylist they still stay for a few cycles (and end with their deatheffect).

I meant to remove them instantly, like clearing effects and effectElements in visualeffects and updating visualeffects manually (to remove the effectNode).

Or to describe it as code:
EffectManagerState
```java
    private void clearActiveEffects() {
        Iterator<VisualEffect> iterator = activeEffects.iterator();
        while (iterator.hasNext()) {
            VisualEffect visualEffect = iterator.next();
            visualEffect.removeEffect();
            visualEffect.update(3.0f);
        }
        activeEffects.clear();
    }
```

VisualEffect
```java
    public void removeEffect() {
        effects.clear();
        effectElements.clear();
        this.infinite = false;
    }
```
This approach is kinda hackish, but you should get the idea

tonihele:
ModelViewer is kinda hackish anyway :-) just a debug tool. But hopefully it
doesnt yield too much hack code to normal business classes we use in game.
I just went with an AppState design but each VisualEffect could also very
well have been a Control. I was just unsure that is it heavier performance
wise. And if we need some centralized control on the effects (like pausing).

ufdada:
> But hopefully it doesnt yield too much hack code to normal business classes we use in game.

That was actually my concern, even if we implement something hackish in our normal classes just for debugging, we might end up using it in our game too (if we need that functionality).

> And if we need some centralized control on the effects (like pausing).

The original one didn't care about pausing effects either, so i don't thing this would be a reason ;)

belgianguy:
Yeah, I'll be careful, hackish workarounds could seep into the main game code and cause bugs or other havoc in the future. But as I don't have a full grasp of the codebase yet, feel free to give pointers and suggestions, they are always welcome!

belgianguy:
> - when switching to something else (like maps) the effect should be removed
- when switching back from maps, the initial marble floor should be seen and the map should be removed

Hmm, I've been reading up on JME, and at the moment I think maybe a separate AppState for Models, Maps, Terrain and Effects would be most applicable in ModelViewer, as they feature their own initialize and cleanup methods and have their own update loop. 

Thoughts?

tonihele:
I think that would be a bit too complex. Switching between model or type
and model just always needs to wipe out the same branch in scene graph. And
there is no functionality needed for separate update loops? It is always
just the rotation? So it would be just the same code duplicated to multiple
states (or inherited).

I can refactor the effects from AppState design to more native Control
design. A small change but then perhaps easier to use. The is the JME that
calls the update loop.

On 6 Aug 2016 17:17, "belgianguy" <notifications@github.com> wrote:


   - when switching to something else (like maps) the effect should be
   removed
   - when switching back from maps, the initial marble floor should be seen
   and the map should be removed

Hmm, I've been reading up on JME, and at the moment I think maybe a
separate AppState for Models, Maps, Terrain and Effects would be most
applicable in ModelViewer, as they feature their own initialize and cleanup
methods and have their own update loop.

Thoughts?

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/tonihele/OpenKeeper/pull/213#issuecomment-238025351>,
or mute the thread

ufdada:
You could do that, but in my opinion it would just be a little bit to much work as it's only for the modelviewer.

My suggestion:
- EffectStateManager should take enabled into account and don't update if the state is not enabled (overwriting setEnabled)
- when disabled, it should remove all ongoing effects
- disable EffectStateManager before the switch case and only enable it on effects
- remove the current loaded spatial with `rootNode.detachChildNamed(NODE_NAME);`
- show the floor with `floorGeom.setCullHint(Spatial.CullHint.Never);`

Instead of enabling it you could attach the EffectStateManger to the StateManager and then detach it if is has that state. Bare in mind you then have to remove the update from simpleUpdate (which i consider bad anyway).

ufdada:
Oh i just noticed setEnabled of EffectManagerState would've worked if it would have been attached to the StateManager in the first place...

ufdada:
I think this is already good enough, i make some small adjustments to it later.

PR #214, opened 2016-08-02
Feature 166 Part 2

belgianguy:
Now the Effects loop correctly, the ModelViewer will always load one effect, and it now uses the update function.


ufdada:
Is there a reason for making a separate pull request?

If you want to update your old pull request you just have to make a commit to its existing branch and it will be automatically updated.

belgianguy:
That was the way I had it, but at that time the scope of the PR wasn't oriented to the feature yet.
I think I have them both together on a branch somewhere. Then I'll close these 2 and just PR the one where they both are. In case that fails, I'll cherry pick the commit from here and put into the 'Part 1' branch.

PR #225, opened 2016-08-27
Implement system messages

### LINKED ISSUE #220: Implement system messages
System messages are used to lead the player throughout the game.

Some features:
- most of them are actually voiced over by the mentor
- they fade in from right to left
- they have a lifetime (about 50 seconds) and then disappear
- they blink if they are unread (info messages in blue, important ones in red), otherwise (hover and read) they just show the white icon
- right click dismisses them (they are removed without animation)
- some have a magnifying glass to move the camera to a specific place, attach it to an object.
- ✘ dismisses them, ✔ just keeps them
- on hover, their text is shown in the tooltip label.
- Huge texts have scrolling

I wasn't able to find out what happens if you get to many of them, like if the first one get deleted or some kind of scrolling is happening

ufdada:
Resolves #220 

This is a basic implementation of the system messages. Currently it only shows the scripted (triggered) speeches of the mentor. I also using now a smaller scrollbar, which fits the theme better than the big one of the main menu.

main1

I found several things that should be fixed to make this work flawlessly:

1. The player interaction needs to be revamped. The gui recognition is more a guess than a correct determination (it tries to figure out the viewport). Therefore the cursor isn't correct and a implementation of the tooltip doesn't make sense because it would overwritten immediately.
2. Focusing things isn't implemented right now (like the first creature appearance which can be overwritten like in the first level)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ local.properties
 
 *.dylib
 /build/
+
+# gradle stuff #
+/private/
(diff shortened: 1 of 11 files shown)


tonihele:
Very nice :)

PR #228, opened 2016-08-28
Fix cursors to support other formats than RGBA 32-bit

ArchDemons:
+ added support BGR8 cursor image format
* fixed idle cursor hotspot

### DIFF
--- a/src/toniarts/openkeeper/gui/Cursor.java
+++ b/src/toniarts/openkeeper/gui/Cursor.java
@@ -19,17 +19,18 @@ package toniarts.openkeeper.gui;
 import com.jme3.asset.AssetManager;
 import com.jme3.cursors.plugins.JmeCursor;
 import com.jme3.texture.Image;
+import com.jme3.texture.Image.Format;
 import com.jme3.texture.Texture;
 import com.jme3.util.BufferUtils;
 import java.io.File;
 import java.nio.ByteBuffer;
 import java.nio.IntBuffer;
-import java.util.regex.Matcher;
 import toniarts.openkeeper.tools.convert.AssetsConverter;
+import toniarts.openkeeper.tools.convert.ConversionUtils;
 
 public class Cursor extends JmeCursor {
 
-    public int delay = 30;
+    public final int delay = 30;
 
     /**
      * Generates a static cursor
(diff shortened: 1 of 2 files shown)


src/toniarts/openkeeper/gui/Cursor.java
```diff
@@ -19,17 +19,18 @@
...
 import java.nio.ByteBuffer;
 import java.nio.IntBuffer;
-import java.util.regex.Matcher;
 import toniarts.openkeeper.tools.convert.AssetsConverter;
+import toniarts.openkeeper.tools.convert.ConversionUtils;
 
 public class Cursor extends JmeCursor {
 
-    public int delay = 30;
+    public final int delay = 30;
```
tonihele: Oooh, my bad, this should be a constant (final static)? No need to change it though, just spotted here :)

tonihele:
Just wondering, has the possession cursor now a transparent background?

src/toniarts/openkeeper/gui/Cursor.java
```diff
@@ -101,4 +101,27 @@ public Cursor(AssetManager assetManager, String Filename, int hotspotx, int hots
...
         this.setImagesData(image);
     }
+
+    private int getARGB(Format format, ByteBuffer data) {
+	int result = 0;
+
+	switch (format) {
+            case BGR8:
+                result = 0xFF000000 | (data.get() & 0xFF) | ((data.get() & 0xFF) << 8) | ((data.get() & 0xFF) << 16);
+                if (result == 0xFF00FF00) { // green only
```
ArchDemons: check if green only color and change it to transparent

src/toniarts/openkeeper/gui/Cursor.java
ArchDemons: all not changeable must be static ! ^_^

tonihele: For clarity & performance!

ArchDemons:
We also have cursors in folder **Textures/cursor**. What we must do with them ?

tonihele:
Hmm, what kind of stuff is there (I'm on a business trip, back on Wednesday evening, no access to resources)?

But other that this PR is ok, and a good job sticker :) Better than my suggestion to fiddle around with BufferedImages. Ok to merge?

ArchDemons:
**Textures/cursor** contain Pointer, idle, press cursors. I think them forgot to remove.

You can merge.

tonihele:
Maybe some old stuff. I also heard that DK II doesn't really use a cursor. Just a quad etc. following the mouse. Maybe they tried to use a cursor with these :)

PR #238, opened 2016-09-11
Keeper Hand

ArchDemons:
* refactor of PlayerInteractionState. Keeper hand in own class
+ added keeper hand item icons queue.


tonihele:
Cool :)

PR #239, opened 2016-09-12
Small Controls Update

ArchDemons:
added FunnyCameraContol (follows the target)
added HighLightControl (marks object. In next update will work properly)
added PossessionCameraControl entrance and exit from possession
fixed Possession selection creature

P.S. FunnyCameraContol is temporary solution (for fun) ^_^

### DIFF
--- a/src/toniarts/openkeeper/game/FunnyCameraContol.java
+++ b/src/toniarts/openkeeper/game/FunnyCameraContol.java
@@ -1,0 +1,234 @@
+/*
+ * Copyright (C) 2014-2016 OpenKeeper
+ *
+ * OpenKeeper is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * OpenKeeper is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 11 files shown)


tonihele:
I'm going to do a bit changes. In IInteractiveControl I had the onHover. I thought that I would make a time based effect. But it maybe easier to -> onHoverStart & onHoverEnd. I watched the original a bit and it is pretty precise. So not timed... So I'm going to make your control the base class for Creature & Object control.

tonihele:
I already made our cursor update every 250ms.

ArchDemons:
IInteractiveControl object updates every update cycle in my version of PlayerInteractionState. And no needed remake them on time based. https://github.com/ArchDemons/Keeper/commit/a1e13e52b391b96612daa7ebd0356353c48235b0#diff-03046a7f8c30284e124b82b86e5bf024R220

tonihele:
Mmm, that was the one merged? For me it doesn't work. It just flickers them in red. And feels a bit more hack. Better to let the controls implement the stuff since there are some factors that affect the decision that should it turn red. Better to be figured out in the object/creature control

ArchDemons:
that does not work, because it is not merged. I will prepare a PR in the near future

tonihele:
Here:
https://github.com/tonihele/OpenKeeper/commit/4a38aea710d11581aed38e157d60c5966dd56d67

The hierarchy is a bit funny now perhaps, but it works.

tonihele:
Ah ok, I misunderstood:

> In next update will work properly

tonihele:
Cool that transition effect on possession, btw

tonihele:
And for the FunnyCamera, there are several cameras in the original game. Not necessarily this 3rd person kind chase cam, but we could still keep and offer it. Perhaps with more descriptive name ;) But it looks nice so why not

ArchDemons:
LockedCamera is most suitable. Need rename others cameras: TopDownCamera and FirstPersonCamera

PR #242, opened 2016-09-15
New PlayerInteractionState

ArchDemons:
refactor PlayerInteractionState
refactor selectionHandler controls visual selection box now

### DIFF
--- a/src/toniarts/openkeeper/view/PlayerInteractionState.java
+++ b/src/toniarts/openkeeper/view/PlayerInteractionState.java
@@ -72,7 +72,7 @@ import toniarts.openkeeper.world.room.RoomInstance;
  * @author ArchDemon
  */
 // TODO: States, now only selection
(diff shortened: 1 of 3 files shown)


tonihele:
Hmm, with this the mouse cursor gets updated every frame? I think that is a bit waste of resources

tonihele:
Also I noticed that when there is something in Keeper's hand while trying to tag land, the selection indicator is RED (as in no, you can't drop the guy here). It should be BLUE. This however happens originally too with our code, so it is not regression. Very small issue, I can fix it later too., just noticed it.

ArchDemons:
> Hmm, with this the mouse cursor gets updated every frame? I think that is a bit waste of resources

Every frame updated flags (onGui, onMap ...). Cursor updates if one of flags is changed. This solution is not ideal, but better than it was. We can build flags tree and not update all flags, only some of them.

For cursor we need param that save old state. And change cursor if new state != old state. E.g. we cans save cursor type.

PR #244, opened 2016-09-16
Sound patch

ArchDemons:
Sound extractor prepared to convert all audio data
added SF2 file format decoder
fixed Wav decode from SDT archive
refactor ConversionUtils
refactor BankMap, SfxMap file decoder (decoded some values)


tonihele:
I always wondered why the few WAV files didn't work. Quite simple reason it seems :) Nice

PR #249, opened 2016-09-29
New steering

ArchDemons:
added faceing creatures
rollback some debug stuff
fixed bug with tile tooltip
fixed creature steering

### DIFF
--- a/src/toniarts/openkeeper/world/creature/CreatureControl.java
+++ b/src/toniarts/openkeeper/world/creature/CreatureControl.java
@@ -23,6 +23,7 @@ import com.badlogic.gdx.ai.steer.behaviors.Cohesion;
 import com.badlogic.gdx.ai.steer.behaviors.FollowPath;
 import com.badlogic.gdx.ai.steer.behaviors.PrioritySteering;
 import com.badlogic.gdx.ai.steer.behaviors.RaycastObstacleAvoidance;
+import com.badlogic.gdx.ai.steer.behaviors.ReachOrientation;
 import com.badlogic.gdx.ai.steer.behaviors.Wander;
 import com.badlogic.gdx.ai.steer.proximities.InfiniteProximity;
 import com.badlogic.gdx.ai.steer.utils.paths.LinePath;
@@ -46,6 +47,9 @@ import toniarts.openkeeper.ai.creature.CreatureState;
 import toniarts.openkeeper.game.action.ActionPoint;
 import toniarts.openkeeper.game.party.Party;
 import toniarts.openkeeper.game.task.AbstractTask;
+import toniarts.openkeeper.game.task.AbstractTileTask;
+import toniarts.openkeeper.game.task.worker.ClaimWallTileTask;
+import toniarts.openkeeper.game.task.worker.DigTileTask;
 import toniarts.openkeeper.gui.CursorFactory;
 import toniarts.openkeeper.gui.CursorFactory.CursorType;
 import toniarts.openkeeper.tools.convert.map.ArtResource;
@@ -64,6 +68,7 @@ import toniarts.openkeeper.world.WorldState;
 import toniarts.openkeeper.world.control.IInteractiveControl;
 import toniarts.openkeeper.world.creature.steering.AbstractCreatureSteeringControl;
 import toniarts.openkeeper.world.creature.steering.CreatureRayCastCollisionDetector;
+import toniarts.openkeeper.world.creature.steering.TargetLocation;
 import toniarts.openkeeper.world.listener.CreatureListener;
 import toniarts.openkeeper.world.object.ObjectControl;
 import toniarts.openkeeper.world.room.GenericRoom;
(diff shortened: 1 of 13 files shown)


tonihele:
Heh, rather comical. The speed is very high now, they miss their targets quite often and spin wildly near target. But otherwise the facing is nice. I think it needs some fixes before merging.

ArchDemons:
we can slowing rotation. But looks that bad. Something wrong with calculation angular acceleration or not limits it. Or we need firstly check acceleration first, than apply them...

ArchDemons:
Hmmm, this looks different on my other computer... How it may happen?

src/toniarts/openkeeper/world/creature/steering/AbstractCreatureSteeringControl.java
```diff
@@ -105,19 +102,27 @@ protected void applySteering(SteeringAcceleration<Vector2> steering, float tpf)
...
+            return;
         }
 
         // Update orientation and angular velocity
-        if (independentFacing) {
-            setOrientation(getOrientation() + (angularVelocity * tpf));
-            angularVelocity += steering.angular * tpf;
+        if (!stopMoving) {
+            position.add(linearVelocity.x * tpf, linearVelocity.y * tpf);
+            linearVelocity.mulAdd(steering.linear, tpf).limit(maxLinearSpeed);
```
ufdada: Are these 2 lines are really necessary? They are getting called some lines above, so if it runs into this block it is called twice.

At least my tests made them move like before (without the weird spinning)

src/toniarts/openkeeper/world/creature/steering/AbstractCreatureSteeringControl.java
```diff
@@ -37,23 +38,27 @@
...
+    private float speedThreshold = 0.1f;
     private volatile boolean applySteering = false;
+    // FIXME: dirty hack. Roll back changes and see what happens :-(
+    private volatile boolean stopMoving = false;
 
     public AbstractCreatureSteeringControl(Creature creature) {
         this.creature = creature;
 
         maxLinearSpeed = creature.getSpeed();
+        maxLinearSpeed = maxLinearSpeed * 10;
```
ArchDemons: `maxLinearAcceleration = maxLinearSpeed * 10;`

but that fix not fixed all bugs :(

ufdada: Look at the other reply ;)

src/toniarts/openkeeper/world/creature/steering/AbstractCreatureSteeringControl.java
ArchDemons: Main idea:
-add FollowPath and Face steering
- creature move to target
- when moving stops (end point is reached), creature rotates around Y-axis (faced).

But on my other computer with downloaded gdx-ai library it feel different with other computer.

src/toniarts/openkeeper/world/creature/CreatureControl.java
```diff
@@ -538,22 +549,42 @@ public short getOwnerId() {
...
-            }
+            // Debug
+//            if (Main.isDebug()) {
+//                worldState.getApp().enqueue(() -> {
+//                    worldState.drawPath(waypoints);
+//                });
+//            }
+
+            FollowPath<Vector2, LinePathParam> followPath = new FollowPath(this, new LinePath<>(getWaypoints(outPath), true), 2);
+            followPath.setDecelerationRadius(0.3f);
```
ArchDemons: Seems like library ignoring this attribute. Or I wrong understood it.

ArchDemons: imps steel cut corners when moves. https://gamedevelopment.tutsplus.com/tutorials/understanding-steering-behaviors-path-following--gamedev-8769 Smoothing the Movement header. Where this attribute in gdx-ai ?

tonihele:
Sorry it took this long, it definitely looks better now. They still easily get stuck though. Like they don't know that they're at the destination already. They just keep the running animation and stay stationary forever. Something with the facing perhaps, since the tasks were digging & claiming a wall when I saw them do this.

The steering behavior parameters can be a bit of pain. I don't have much time on the weekend, but I can also try to debug it. It is really delicate fine tuning perhaps.

tonihele:
This is _impossible_ :( I tried cherry picking the facing, just the facing. And I added collision avoidance system. They all work.. to a degree... It is coming painfully obvious that the system needs to be customized more (See: https://github.com/jsjolund/GdxDemo3D/blob/master/core/src/com/mygdx/game/steerers/FollowPathSteerer.java). I disabled them for now. Changing any parameter yields too random results. We need better understanding.

We also need a bit better hierarchy, more controlled structure for clarity. Proper MVC. As the code keeps growing. Inheriting the AbstractCreatureSteeringControl is not perhaps the best design. Somewhat related matter to this steering.

ufdada:
Superseded by #269

PR #250, opened 2016-09-29
Flow layout

ArchDemons:


src/toniarts/openkeeper/gui/nifty/flowlayout/CustomScroll.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public void back() {
+        // FIXME not worked
+        content.setConstraintX(SizeValue.px(content.getConstraintX().getValueAsInt(1.f) - 64));
+    }
+
+    public void forward() {
+        // FIXME not worked
+        content.setConstraintX(SizeValue.px(content.getConstraintX().getValueAsInt(1.f) + 64));
```
tonihele: If these didn't work, then you might need to call that layoutElements(), on parent.

ArchDemons: Problem not only in this. i read documentation and find how fix it.

ArchDemons: Problem not only in this. i read documentation and find how fix it.

tonihele:
Looks nice but I could get it to work?
flow

tonihele:
It leaves the last item unreachable. But anyway it is good. We can hone it later. At the same time we can implement the big icons. In DK II it uses large icons when there is only few items to show. We have bigger resolutions available so especially we would benefit from implementing this behavior. Makes it easier to use. Big icons.

tonihele:
And also that Nifty double click thing is a bit annoying. I've seen it before somewhere that the single clicking interval must be very slow. But here it really shows. I wonder if there is another solution than to try to hook it like onDoubleClick="forwardTwice" etc.

tonihele:
https://github.com/nifty-gui/nifty-gui/issues/247

PR #255, opened 2016-10-23
Effect 2

ArchDemons:


src/toniarts/openkeeper/view/KeeperHand.java
```diff
@@ -188,7 +189,9 @@ private void removeIcon() {
     }
 
     private void addIcon(final ArtResource image, final int index) {
-        if (!(image.getSettings() instanceof ArtResource.Image)) {
+        if (image.getType() != ArtResourceType.ALPHA
+                || image.getType() != ArtResourceType.ADDITIVE_ALPHA
```
ufdada: I think there should be an && operator

src/toniarts/openkeeper/view/KeeperHand.java
```diff
@@ -188,7 +189,9 @@ private void removeIcon() {
...
+                || image.getType() != ArtResourceType.SPRITE) {
```
ufdada: and here

ArchDemons:
Known bug with torch flame. It's not in correct position (I do not know how to fix it). Need change Material of Flame from current to Particle (problem with background black color), but when I change, I got black rectangle...

tonihele:
It is nice though, our torches aren't yet in right places anyway :) Only troubles me that the overall atmosphere in our dungeon doesn't reflect the original :) I'll try to look it closer tomorrow, I tried it already briefly yesterday.

We have a fundamental problem with the placing. We need to reset all models somehow, feels wrong, maybe a simple thing but I never studied it further. Maybe causing all sorts of problems down the line.

src/toniarts/openkeeper/tools/convert/map/Trigger.java
```diff
@@ -32,38 +32,38 @@
...
+            data = new HashMap<>();
         }
 
-        if (data == null) {
-            userData.remove(key);
-        } else if (data instanceof Number) {
-            userData.put(key, data);
+        if (value == null) {
+            data.remove(key);
+        } else if (value instanceof Number) {
```
tonihele: Same here?

src/toniarts/openkeeper/tools/convert/map/ArtResource.java
```diff
@@ -93,219 +100,63 @@ protected void setName(String name) {
...
-            return id;
+    public Collection<String> getDataKeys() {
+        if (data != null) {
+            return data.keySet();
         }
 
-        protected void setId(int id) {
-            this.id = id;
-        }
+        return Collections.EMPTY_SET;
```
tonihele: I believe one should use the methods instead, emptySet()

src/toniarts/openkeeper/tools/convert/map/Trigger.java
ArchDemons: it`s always number

tonihele:
Nice nice

PR #257, opened 2016-11-13
Creature messages

ufdada:
This add system messages when a creature enters for the first time the dungeon.

If the introduction override is set, it will display a custom message and play the corresponding mentor speech (like for the goblin in level 1).

### DIFF
--- a/src/toniarts/openkeeper/view/PlayerCameraState.java
+++ b/src/toniarts/openkeeper/view/PlayerCameraState.java
@@ -212,6 +212,7 @@ public class PlayerCameraState extends AbstractPauseAwareState implements Action
             @Override
             public void onPlay(CinematicEvent cinematic) {
                 stateManager.getState(PlayerState.class).setTransitionEnd(false);
+                stateManager.getState(PlayerState.class).setWideScreen(true);
                 inputManager.setCursorVisible(false);
                 PlayerCameraState.this.cameraStore();
             }
@@ -223,6 +224,7 @@ public class PlayerCameraState extends AbstractPauseAwareState implements Action
             @Override
             public void onStop(CinematicEvent cinematic) {
                 stateManager.getState(PlayerState.class).setTransitionEnd(true);
+                stateManager.getState(PlayerState.class).setWideScreen(false);
                 inputManager.setCursorVisible(true);
                 PlayerCameraState.this.cameraRestore();
             }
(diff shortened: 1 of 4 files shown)


src/toniarts/openkeeper/world/WorldState.java
```diff
@@ -188,6 +191,26 @@ private void initPlayerCreatures() {
 
                 @Override
                 public void onSpawn(CreatureControl creature) {
+                    if (player.getId() == stateManager.getState(PlayerState.class).getPlayerId() && player.getCreatureControl().getTypeCount(creature.getCreature()) == 0) {
```
tonihele: This will do for now, but in the long run this doesn't work. As they get killed or sacked. Easy enough to track this per level, but how about on campaign (it doesn't show this on Goblins every time)? Is there some parameters on the level data?

No need to solve this now though, this is great just as it is. Just starting a conversation :)

tonihele: Yes, that can be added to the PlayerCreatureControl since:
 - We actually should remove the entry when count == 0, so we can't use contains() to check this. Because of the UI ordering. The UI doesn't show 0 and I think it shows them in the order which the types spawned, minor detail

We should check how the original does the intros. My assumptions:
 - In campaign levels you get intro only for the totally new creature types -> if so, this could be as simple as checking the level type and if there is an override -> show intro, otherwise no
 - In skirmish/mp/mpd levels you always get the intro for the first time a type is introduced

tonihele: ^ We need an issue for this

src/toniarts/openkeeper/view/PlayerCameraState.java
```diff
@@ -212,6 +212,7 @@ public void doTransition(int sweepFileId, final ActionPoint point) {
             @Override
             public void onPlay(CinematicEvent cinematic) {
                 stateManager.getState(PlayerState.class).setTransitionEnd(false);
+                stateManager.getState(PlayerState.class).setWideScreen(true);
```
ArchDemons: we have differently trigger with widescreen action. Not shure that we need add this action to transition

ufdada: As far as i can tell this is the case in the original game as well. For example the goblin intro in level1, there is also the cinematic screen but there is no trigger for setting it to widescreen.

PR #265, opened 2017-01-06
Implement basic seeing & reacting to each other

tonihele:
Also stun, and some other improvements etc. Doesn't really solve anything yet, but doesn't break either.


ufdada:
Level 1 becomes nearly playable 👍 😁 . But the goblins are rather fleeing than fighting and the Hero takes one by one :(

tonihele:
I'm actually trying to make level 1 just now playable, and making those dwarfs dig & enemies attack. The fleeing part was difficult, I don't fully understand the factors in it. It is not simply threat > fear. Feel free to tweak with it, there is also the health critical. I'll upload my "Arena", it'll save 5min of map making time :)

PR #266, opened 2017-01-17
Bugfix

ArchDemons:


src/toniarts/openkeeper/Main.java
```diff
@@ -360,7 +360,7 @@ public Void onLoad() {
                     List<String> guiXMLs = Arrays.asList("Interface/MainMenu.xml", "Interface/GameHUD.xml");
                     for (String xml : guiXMLs) {
                         try {
-//                            nifty.validateXml(xml); <-- Amazingly buggy?
+                            nifty.validateXml(xml);
```
tonihele: Nice that you figured this one out

tonihele:
I just have to remember to add the diagonal movement to the interface (branch level1). Digging is not allowed diagonally that is...

ArchDemons:
Not fixed bug with dropping creature. After this action they navigate to random point from tile where they picked up

src/toniarts/openkeeper/world/creature/CreatureLoader.java
```diff
@@ -108,9 +108,9 @@ public String getCenterIcon() {
 
     public static void setPosition(Spatial creature, Vector2f position) {
         creature.setLocalTranslation(
-                position.x * MapLoader.TILE_WIDTH - MapLoader.TILE_WIDTH / 2f,
+                MapLoader.TILE_WIDTH * (position.x - 0.5f),
```
tonihele: Isn't this formula basically wrong? If our WIDTH isn't 1, this doesn't work at all?

tonihele:
I think it worked before, is it the new Landing control?

tonihele:
Ah yes, it must be because the position is actually on the model. On drop it should be set again. As the visual presentation is briefly the master there.

tonihele:
Hmm, but it can't be it, it should just warp then

src/toniarts/openkeeper/animation/PoseTrack.java
```diff
@@ -170,7 +171,11 @@ private void applyFrame(Mesh target, int frameIndex) {
      * @param weight weight on which to apply the interpolation
      * @param vertexBuffer the vertex buffer
      */
-    private void applyPose(Pose startPose, Pose endPose, float weight, FloatBuffer vertexBuffer) {
+    private void applyPose(@Nullable Pose startPose, Pose endPose, float weight, FloatBuffer vertexBuffer) {
+        if (startPose == null) {
+            // FIXME should we skip if null?
```
tonihele: Hmm, I'm curious about this. Is there some problems with our animations?

ArchDemons: sometimes. Maybe when debug and pause on some thread

src/toniarts/openkeeper/world/creature/CreatureLoader.java
ArchDemons: a * b - b * c = b * (a - c) mathematic ^_^. Old and new formulas are correctly.

tonihele: I think you are right :)

tonihele:
Actually I was right. I just didn't try it, just tried to see from code... Anyway it now calls the CreatureLoader, it sets the position also, so it is not the best method, but since all is done within a frame (I think) you'll never see it blinking. But the reason was the position in the model was the old one.

BTW, creatures have mass. And the level variables have gravity.. Should we try that does it affect the creature dropping in the original game? :) Anyway, nice work!

PR #267, opened 2017-01-22
Level1

tonihele:
Basically makes the campaign now progressable. You can win many campaign levels, well, depending on the triggers. But at least the first one. Mainly I want to merge this now since it fixes the remainder of very annoying concurrency bugs on the scene graph. And I'm also leaving for 2 weeks. So if somebody wants to tweak something with this as base, it is now possible. This doesn't break a thing, at least to my knowledge.

KillPlayer objective is implemented, it is not perfect, but it does the job. Objects are not attackable now so you can't really loose by loosing your DHeart.


ufdada:
Great changes! And come back in one piece 😉

PR #269, opened 2017-02-02
New Steering test two

ArchDemons:
added ReachOrientation steering (now work better than was)

### DIFF
--- a/src/toniarts/openkeeper/world/creature/CreatureControl.java
+++ b/src/toniarts/openkeeper/world/creature/CreatureControl.java
@@ -912,7 +912,7 @@ public abstract class CreatureControl extends AbstractCreatureSteeringControl im
      */
     public boolean followTarget(CreatureControl target) {
         if (target != null) {
-            PrioritySteering<Vector2> prioritySteering = new PrioritySteering(this, 0.0001f);
+            PrioritySteering<Vector2> prioritySteering = new PrioritySteering(this);
 
             // Create proximity
             Array<CreatureControl> creatures;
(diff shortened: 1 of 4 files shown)


ArchDemons:
previous Pull Request we can close, it`s not actual.

ufdada:
Seems like the facing of the hero party or the goblins doesn't work with these changes. At least in level 1.

tonihele:
You mean while in battle? That is not implemented before nor in this PR. But it'll be easy to add such if this works as intended. My collision avoidance caused Antonius' (level 1) patrol route errors, but patrol routes do not have target facing anyway. Just a cautionary tale on how difficult the steering is to tweak...

ufdada:
No i mean when antonius is on patrol and the goblin are simply walking. They are always facing in the same direction

tonihele:
:( God dammit steering lib. Y u no work....

ArchDemons:
> Seems like the facing of the hero party or the goblins doesn't work with these changes. At least in level 1.

I test on custom level with imps. But my changes should not  change direction if creature moving. Only if it stay and have angular velocity. Need some debug

ArchDemons:
Problem was too large zeroLinearSpeedThreshold (more than max linear speed)

tonihele:
I think it works in general. Only thing I wound was that when the creatures start moving, they don't know which way to orientate at first. They do this quick "dance move". Clearly visible on Antonius patrolling. But I'd rather have that than no orientation. It is so minor, given that the rest of the movement stuff aren't perfect either.

tonihele:
Nice work

Issue #263, opened 2016-12-26
Refactor to proper MVC design


tonihele:
Refactor to proper MVC design. Or entity design, it is quite the same also. We are now marrying our M with C. Separate them. And indeed the Paul's (pspeed) ES system might be of use. I think we already have it imported, it might provide some automatic stuff for the network play, And there are some tools that might be better than our hand crafted ones in the package. I haven't looked the examples much but no over-designing! Not every little thing needs to be an entity IMHO.

This will help clean out the code and at the same time it is clearer to implement the creature/trap/object actions. They are now bound to the animations, instead they should be just i.e. damage per second. I don't think the animations are synced really, just to sounds. I don't know how the fighting works in that aspect.

tonihele:
When I started I just wanted to display some graphics, to see if it was possible. Well, it is. Now with the added logic, the design doesn't really support many needed features and is no longer expandable without this redesign.

Goal is to decouple the logic from the graphics. Enable network play and save + load game state. Much like JSettlers have done (I haven't actually looked in to their code, just skimmed the packages they have). Here is also what mr. PSpeed thinks as one possible proper architecture: http://simsilica.github.io/SimEthereal/.

This is quite a huge undertaking but it can't wait much longer.

belgianguy:
pspeed has some other examples as well: https://github.com/Simsilica/Examples

tonihele:
Yeah, from what I can gather Ethreal & Zay-ES are smart to marry together

belgianguy:
True, might be interesting to review pspeed's AsteroidPanic (https://github.com/jMonkeyEngine-Contributions/zay-es/tree/master/examples/AsteroidPanic) as well then. (Found it mentioned here: https://hub.jmonkeyengine.org/t/to-entity-or-not-to-entity/37714, which is another dicussion related to mixed code and graphics)

tonihele:
https://github.com/jMonkeyEngine-Contributions/zay-es/tree/master/examples/MonkeyTrap seems better example out of the two (vs. AsteroidPanic). Just looking at the code. Structurewise...

tonihele:
I was thinking... I'm not treating map as any entity etc. And rooms were really my concern. They too would not fit to any entity. If synced over the network they would need to go with the same tileUpdate stuff to be synced in relation to map state. Feels kinda clumsy. Not maybe that hard.

[...]

PR #271, opened 2017-02-28
New MapLoader

ArchDemons:


tonihele:
A lot to take in :D But overall seems like a job well done, thanks. However, I found your lack of water & lava and floating door locks disturbing... I just gave it a quick spin. The cinematic seems to be changed a bit? Visually I mean, the camera is a bit lower? I don't know if this is a fault or not, just commenting, or am I imagining this...?

Did this fix some of the room builds? At least the temple in one of the campaign levels looked better than before.

The door locks I remember were really tricky :( Maybe you fixed something and now my hacks wont work.

ArchDemons:
Today I also found bug with water and lava. Tomorrow push fixes. 
Because map was higher on 1f, camera was lower on 1f. Maybe camera settings change it back (we still do not know how to configure it properly).
I fix DoubleQuad constructor and all rooms on it. 
I don`t check  door locks. Check it tomorrow. Problem maybe with lock height position as water and lava.

ArchDemons:
I fixed all what found

src/toniarts/openkeeper/tools/convert/map/KwdFile.java
```diff
@@ -380,7 +381,7 @@ private void readFileContents(KwdHeader header, RandomAccessFile data) throws IO
     }
 
     /**
-     * Reads the *Map.kld
+     * Reads the *GameMap.kld
```
tonihele: No :) You search & replaced too much :)

src/toniarts/openkeeper/tools/convert/map/KwdFile.java
```diff
@@ -3167,7 +3168,7 @@ private void checkOffset(long itemSize, RandomAccessFile file, long offset) thro
         private int checkThree; // in level = 223
         private int dataSizeLevel; // in Level size of data exclude paths
         private int unknown; // only in Triggers and Level
-        private int headerEndOffset; // 28, *Map - 8, *Triggers - 32,
+        private int headerEndOffset; // 28, *GameMap - 8, *Triggers - 32,
```
ufdada: This shouldn't be renamed, too :)

tonihele:
Nice work sir! I found a small bug, but I'll just fix it after the merge.

tonihele:
Oh and a nice to see the door somewhat working :)

tonihele:
Indeed the camera is too low. You can see it from the creature entrance video. I'm not going to touch it, but you can fix it if you want. I have a lot to do now, putting all this back to #263 :)

ArchDemons:
ok

ArchDemons:
I think we should stay this changes as is. We should not roll back this changes, until we understand what wrong: camera height or map height. Why? If you see torches height in rooms (in editor), you see 0.5f or 1.5f. In current state all good, but if we move map on y=-1f torch position not will not match

tonihele:
Might also be things like origin of the models, or origin on the screen coordinates on view port (can't be...? we are not using any such thing?). Since I think we are converting from DirectX to OpenGL

tonihele:
And the change is good as it is, and also the fact that I'm reaallllly trying to redesign our architecture now, so I appreciate that there would be no super major changes that are hard to pick to new architecture (and might even be not relevant anymore). I'll try to be quick with it, rather than try to implement a lot of new stuff at the same time, like I already actually tried (player maps, but our current map rendering is actually quite difficult, more specifically how we handle the rooms, the map loader needs to know too much just for rendering purposes)...

PR #273, opened 2017-03-22
Modelviewer

ArchDemons:


tonihele:
Nice by the codewise looks of it. This actually helps too my refactoring. At one point I was accepting the fact that ModelViewer is gone at the same time I separate our stuff, I just can't support it. Looks interesting those Creature thingies, need to try it...

tonihele:
Nice, especially secret door is erotic to watch :)

src/toniarts/openkeeper/tools/modelviewer/nifty/CreatureControl.java
```diff
@@ -0,0 +1,62 @@
...
+import de.lessvoid.nifty.controls.Parameters;
+import de.lessvoid.nifty.elements.Element;
+import de.lessvoid.nifty.input.NiftyInputEvent;
+import de.lessvoid.nifty.screen.Screen;
+
+/**
+ *
+ * @author archdemon
+ */
+public class CreatureControl implements Controller {
```
tonihele: This is not used? Maybe better delete

ArchDemons: For future use

tonihele: Lol, ok, I'll take it in as a baby you need to nurture in the future :)

Issue #223, opened 2016-08-21
Implement creature picking up to creature cards


tonihele:
Implement creature picking up to creature cards. Similar to the going through them when zooming. And also they aren't supposed to be counted while in hand.

PR #276, opened 2017-04-09
Pickup from card

ArchDemons:
we need creatures population calculator like workers population calculator

### DIFF
--- a/assets/Interface/GameHUD.xml
+++ b/assets/Interface/GameHUD.xml
@@ -275,19 +275,18 @@
                         <!--The rest-->
                         <control name="customTabGroup" height="100%" width="100%" valign="top" id="tabs-hud">
 
-                            <control name="customTab" id="tab-creature" image="Textures/GUI/Tabs/t-creatures.png"
-                                     active="Textures/GUI/Icons/selected-creature.png" hintText="${menu.2153}" tooltip="${menu.2154}">
+                            <control name="customTab" id="tab-creature"
+                                     image="Textures/GUI/Tabs/t-creatures.png"
+                                     active="Textures/GUI/Icons/selected-creature.png"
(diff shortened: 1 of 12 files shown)


ufdada:
Resolves #223 .

Looks good to me.

There is a slight delay when trying to pickup several minions very fast, but it seems it is because of our state design (it gets executed every tick)

ArchDemons:
nifty interprets fasts clicks as double click. And that is another handler

ufdada:
That makes sence :).
Can we attach the creature pickup to the dblclick handler? It would picking up a few creatures less tricky.

tonihele:
See:
https://github.com/tonihele/OpenKeeper/pull/250

I think the same is on zooming on the cards as well...

tonihele:
I don't think we are utilizing the double click anywhere? Should check the code. If not, just set the delay to 100ms or even lower

ArchDemons:
If we can lower delay between click we should do that

tonihele:
```
nifty.setGlobalProperties(new Properties());
nifty.getGlobalProperties().setProperty("MULTI_CLICK_TIME", "100");

```
?

ArchDemons:
nifty.getGlobalProperties().setProperty("MULTI_CLICK_TIME", "1");

:)

tonihele:
If we aren't using double click anywhere, why not :) Can you test?

ArchDemons:
this solution works. But where I should put it (what class)? Main -> simpleInitApp() ?

tonihele:
https://github.com/tonihele/OpenKeeper/blob/master/src/toniarts/openkeeper/Main.java#L357 ? Yeah, might be most logical

tonihele:
Great thanks! Imps zoomTo doesn't work though.

ArchDemons:
fix zoomTo in next commit

PR #277, opened 2017-04-15
Fixed some bugs

ArchDemons:
fixed interface merge rollback
fixed low cinematic position in some case
fixed bug with tile health
fixed bug with visibles creatures in debug mode
fixed bug with room tooltip
some rooms can spawn creatures now


tonihele:
Is it really that the portals (or others) spawned count is room specific? I'm away and on Linux currently. The current spawning controller doesn't work like that. It is always total. So if it is room specific, then it also needs to be changed.

tonihele:
The combat doesn't work so "well" anymore. The navigation to an opponent is now somehow worse.

src/toniarts/openkeeper/world/TileData.java
```diff
@@ -216,7 +216,7 @@ public boolean applyDamage(int damage) {
      * @return true if the tile is at max
      */
     public boolean applyHealing(int healing) {
-        health = Math.min(getTerrain().getMaxHealth(), health + healing);
+        health = (int) Math.min(getTerrain().getMaxHealth(), (long) health + healing);
```
tonihele: Hmm, why the casts are added?

src/toniarts/openkeeper/world/room/GenericRoom.java
```diff
@@ -618,4 +602,43 @@ public GameObject getPillarObject() {
...
+        temp.put(6, "%48"); // Library
+        temp.put(7, "%49"); // Training Room
+        temp.put(10, "%50"); // Workshop
+        temp.put(9, "%51"); // Guard Rooms
+        temp.put(16, "%53"); // Combat Pit
+        temp.put(12, "%54"); // Torture
+        temp.put(11, "%58"); // Prison
+        temp.put(14, "%61"); // Graveyard
+        temp.put(13, "%62"); // Temple
+        temp.put(15, "%63"); // Casino
```
tonihele: Not very good, but if you need to do this, at least make the hash static instead of always creating a new object.

tonihele:
Sorry for the interface rollback. Nowadays we also use GIT at work, but still manage to f* things up :) I should try SourceTree. It seems better that the NetBeans build in one.

src/toniarts/openkeeper/world/TileData.java
ArchDemons: after claiming the portal we `applyHealing(Integer.MAX_VALUE);`. What happen if we make `(int) 10 + Integer.MAX_VALUE` ? Ea -2431254325 (some big negative value).

`Math.min(getTerrain().getMaxHealth(), health + healing);` return this negative value without cast.

tonihele: :D Okay, that is why it was so weird percentage on the tooltip.

ArchDemons:
About portal attracted. When I made this, I did not have the game to check how need.

About fight. In our old solution creature take target I does not see anything more. It run throw rock to target. My solution not ideal. And someday we fix. But not now.

src/toniarts/openkeeper/world/room/GenericRoom.java
ArchDemons: I thing we move this part to each room class with overriding `getTooltip()`

tonihele:
I think I checked when I made it, but it was a long time ago. We should check. Since now with this it is more messy. And leaves the references forever in the rooms (not an issue right now if you are not on a killer debug session :) ). I think they are collected from the totals.

You mean that the target runs faster and we don't see it anymore but we are in pursue? I'm a bit reluctant to accept this change to fighting, although the old was not perfect, but it was more playable. With this it takes forever to get them to fight if it is ever successful.

ArchDemons:
problem current fight system load Antonius take target Imp and seek him until it was killed. Antonius not see goblins, he did not care about this. Never mind  that 5 goblins are attacking me. The main target - imp :)

ArchDemons:
I can not work with our state machine. Maybe someday ... At current time it blows my brain

tonihele:
Hmm, he shouldn't chase the imp then. Hmm, I tried to make it so that if target flees, attacker doesn't pursue if other targets are visible. If I remember correctly there was a variable in the game variables that how long we should pursue. But there should be a different state for that.

Yes, it is difficult. Easy to break :( But it too will have to go to trash on my redesign.

tonihele:
if (attackTarget == null || attackTarget.isIncapacitated() || attackTarget.getStateMachine().getCurrentState() == CreatureState.FLEE), hmm, it should not chase the imp...

ArchDemons:
But we attach seek steering. And you solution is not correct creature must attack FLEE target. But it can not attack if not see target. 

1. if killer hear target, it should move to victim (must see target).
2. if killer see target, it must run to it (it last position) on attack distance
3. when killer run to victim it must check more relevant target (target on attack distance)
4. if killer on attack distance - fight until somebody DEAD or FLEE.
5. if target flee out of hearing (visible) distance, killer change state to IDLE

tonihele:
Can't it just check the visibility then? If the enemy is visible, you seek rather than go to its last known position. This is in real life too :) Now it indeed doesn't check the visibility.

Currently the hearing & seeing isn't separated, I don't even know was this implemented in the original. For visibility & hearing everybody has 4 of range.

It would be easy to just modify the getAttackTarget() method so that it would check the visibility. And if the current target is FLEEing, just try to find more reasonable target, if nothing found, pursue the FLEEing one.

For getting more advanced behavior, this is what you probably seek, that if there is no one to attack anymore, but only due to lost of visibility -> use use navigation to go to last seen position (this needs to be added to variable). Put the state to that MOVE. It will attack if presented the opportunity, right?

ArchDemons:
our current realization or visibility like hearing. And if killer seek target it move throw rock. This is not right.

You can see in original that good creature stop moving to kill target and continue last work if target move away visibility or hearing zone.

tonihele:
> our current realization or visibility like hearing. And if killer seek target it move throw rock. This is not right.

Can you rephrase this ? I think I don't understand what you mean by this.

So you don't see them ever pursuing a fleeing target?

ArchDemons:
I saw how dwarf moves through rock tile when seek imp. Because imp flee from dwarf

tonihele:
Yes, that will happen. Less likely with the navigation than seek. But we have no prevention for that currently. They still don't see through rock, or hear. But they do cut the corners. And if you have corridors rather than open spaces, they do clip the wall.

ArchDemons:
no no no. Not like cut the corners.

tonihele:
I understand what you mean, they literally go through the rock, yes yes. And with navigation they do it also. But less so because they try to follow the path.

tonihele:
For me, both look erroneous, seek & path when attacking. We just need to implement the collision detection. I tried to do it earlier, but the Steering library is some god damn delegate system, that small touch and Lord Antonius flies to another galaxy instead of patrolling his domain.

ArchDemons:
https://realtimeboard.com/app/board/o9J_k09qbMQ=/

tonihele:
Could a good compromise be then for you to tweak the path when attacking so that the battle actually happens? To update the path, and/or put the Path following to priority steering as first, seeking second (might not work...)...?

tonihele:
You created a board? Invite me in, the link says not authorized.

belgianguy:
In the original, Delaunay triangulation was used for both the wonky walls and the pathfinding, but that is pretty much hard math to the extreme.

ArchDemons:
ready for merge

tonihele:
Okay, in that way the path find is correct, that it should be used when we **hear**, but when wee **see** it should be the pursuit behavior. Since even with the collision detection the pursuit may never lead to the target. But since we now don't really separate these two senses.... I guess it is ok... for now.

What comes to the room capacities. Portals don't monitor this on per room basis. I'm quite sure. They don't have capacity at all. Graveyards and prisons do have capacity per room which they do seem to report in the tooltip. But for the end result, skeletons and vampires I think they are monitored as the portals. Globally. So tracking the spawned stuff per room is not right.

The tooltip data is actually quite tricky... In that sense that it has more data than the room entity ever should has access to. Given the structure in which they are in the translations, the % -substitutions are unique. That leads to some central game object that can provide all of these, i.e. TooltipProvider which has access to all data and parses the string for substitutions. We need some thinking on this. For design clarity we shouldn't allow all-to-all access so to speak...

PR #278, opened 2017-04-15
wrapped file path with ConversionUtils.getCanonicalAssetKey

belgianguy:


tonihele:
This is probably ok, and doesn't cause any merge conflicts with @ArchDemons PR

PR #279, opened 2017-04-17
Added minimum fitting square calculation to be able to determine if t…

belgianguy:
…he Temple hand should be drawn

Now the temple coordinatesMatrix gets checked to see if it has a 5x5 square available to its type, if so, the hand is allowed to be drawn. We should still test what happens when we have multiples of 5 (eg a Temple of 15 x 5)

I think we might be able to use this also for calculating the water tiles in the Temple.

### DIFF
--- a/src/toniarts/openkeeper/utils/Dimension.java
+++ b/src/toniarts/openkeeper/utils/Dimension.java
@@ -1,0 +1,39 @@
+package toniarts.openkeeper.utils;
+
+/**
+ * Created by wietse on 17/04/17.
+ */
+public class Dimension {
+    private int x;
+    private int y;
+
+    public Dimension() {
+        this.x = 0;
+        this.y = 0;
+    }
+
+    public Dimension(int x, int y) {
+        this.x = x;
+        this.y = y;
+    }
+
+    public int getX() {
+        return x;
+    }
+
+    public void setX(int x) {
+        this.x = x;
+    }
+
+    public int getY() {
+        return y;
+    }
+
+    public void setY(int y) {
+        this.y = y;
+    }
+
+    public int getArea() {
+        return x * y;
+    }
+}
--- a/src/toniarts/openkeeper/utils/Dimension.java
+++ b/src/toniarts/openkeeper/utils/Dimension.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (C) 2014-2017 OpenKeeper
+ *
+ * OpenKeeper is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * OpenKeeper is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with OpenKeeper.  If not, see <http://www.gnu.org/licenses/>.
+ */
 package toniarts.openkeeper.utils;
 
 /**
--- a/src/toniarts/openkeeper/utils/RoomUtils.java
+++ b/src/toniarts/openkeeper/utils/RoomUtils.java
@@ -68,6 +68,36 @@ public class RoomUtils {
 
     }
 
+    public static boolean matrixContainsRectangle(boolean [][] coordinatesMatrix, int rectangleSideWidth, int rectangleSideHeight) {
+        final int rows = coordinatesMatrix.length;
+        final int cols = coordinatesMatrix[0].length;
+
+        //Calculate F (1 histogram per row)
+        int[][] F = new int[rows][cols];
+        for(int i = 0; i < rows; ++i) {
+            for(int j = 0; j < cols; ++j) {
+                if(coordinatesMatrix[i][j]) {
+                    if(i > 0) {
+                        F[i][j] = 1 + F[i - 1][j];
+                    } else {
+                        F[i][j] = 1;
+                    }
+                } else if(!coordinatesMatrix[i][j])  {
+                    F[i][j] = 0;
+                }
+            }
+        }
+
+        for(int i = rows - 1; i >= 0; --i) {
+            if((maxArea(F[i]).getX() >= rectangleSideWidth && maxArea(F[i]).getY() >= rectangleSideHeight) ||
+                    (maxArea(F[i]).getX() >= rectangleSideHeight && maxArea(F[i]).getY() >= rectangleSideWidth)) {
+                return true;
+            }
+        }
+        return false;
+
+    }
+
     /**
      * @param histogram histogram detailing the amount of sequential room tiles
      * @return the x and y coordinates of the largest fitting rectangle up to this point in the matrix
@@ -133,7 +163,7 @@ public class RoomUtils {
 
         for(int i = 0; i < rows; ++i) {
             for (int j = 0; j < cols; ++j) {
-                waterTiles[i][j] = !isBorderTile(coordinatesAsMatrix, i, j, rows, cols);
+                waterTiles[i][j] = isWaterTile(coordinatesAsMatrix, i, j, rows, cols);
             }
         }
         return waterTiles;
@@ -145,16 +175,16 @@ public class RoomUtils {
      * @param col current column
      * @param rows total rows
      * @param cols total columns
-     * @return whether the current tile is a border tile or not
+     * @return whether the current tile is a water tile or not
      */
-    private static boolean isBorderTile(boolean[][] coordinatesAsMatrix, int row, int col, int rows, int cols) {
-        if(row == 0 || col == 0 || row == rows - 1 || col == cols - 1 ) {
-            return true;
+    private static boolean isWaterTile(boolean[][] coordinatesAsMatrix, int row, int col, int rows, int cols) {
+        if(row <= 0 || col <= 0 || row >= rows - 1 || col >= cols - 1 || !coordinatesAsMatrix[row][col]) {
+            return false;
         }
 
-        return (   !coordinatesAsMatrix[row - 1][col - 1] || !coordinatesAsMatrix[row - 1][col] || !coordinatesAsMatrix[row - 1][col + 1]
-                || !coordinatesAsMatrix[row][col - 1] || !coordinatesAsMatrix[row][col + 1]
-                || !coordinatesAsMatrix[row + 1][col - 1] || !coordinatesAsMatrix[row + 1][col] || !coordinatesAsMatrix[row + 1][col + 1]);
+        return  (coordinatesAsMatrix[row - 1][col - 1] && coordinatesAsMatrix[row - 1][col] && coordinatesAsMatrix[row - 1][col + 1]
+                && coordinatesAsMatrix[row][col - 1] && coordinatesAsMatrix[row][col] && coordinatesAsMatrix[row][col + 1]
+                && coordinatesAsMatrix[row + 1][col - 1] && coordinatesAsMatrix[row + 1][col] && coordinatesAsMatrix[row + 1][col + 1]);
     }
 
     public static boolean[][] calculateBorderArea(boolean[][] coordinatesAsMatrix, boolean[][] waterArea) {
@@ -163,10 +193,18 @@ public class RoomUtils {
 
         boolean[][] borderTiles = new boolean[rows][cols];
 
-        for(int i = 0; i < rows; ++i) {
-            for (int j = 0; j < cols; ++j) {
-                borderTiles[i][j] = !waterArea[i][j] && coordinatesAsMatrix[i][j];
+        for(int row = 0; row < rows; ++row) {
+            for (int col = 0; col < cols; ++col) {
+                borderTiles[row][col] = coordinatesAsMatrix[row][col] && !isWaterTile(coordinatesAsMatrix,row, col , rows, cols) && //special!
 
+                        (isWaterTile(coordinatesAsMatrix,row - 1, col - 1, rows, cols) ||                                                                               isWaterTile(coordinatesAsMatrix,row - 1, col, rows, cols) ||                                                                                        isWaterTile(coordinatesAsMatrix,row - 1, col + 1, rows, cols) ||
+
+                         isWaterTile(coordinatesAsMatrix,row, col - 1, rows, cols) ||                                                                 //                       isWaterTile(coordinatesAsMatrix,row, col, rows, cols) ||
+                         isWaterTile(coordinatesAsMatrix,row, col + 1, rows, cols) ||
+
+                        isWaterTile(coordinatesAsMatrix,row + 1, col - 1, rows, cols) ||
+                        isWaterTile(coordinatesAsMatrix,row + 1, col, rows, cols) ||
+                        isWaterTile(coordinatesAsMatrix,row + 1, col + 1, rows, cols));
             }
         }
         return borderTiles;
--- a/src/toniarts/openkeeper/utils/RoomUtils.java
+++ b/src/toniarts/openkeeper/utils/RoomUtils.java
@@ -1,0 +1,100 @@
+package toniarts.openkeeper.utils;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * Created by wietse on 17/04/17.
+ */
+public class RoomUtils {
+
+    /*
+    coordinatesMatrix: TRUE means of part of the room type, FALSE means not part of the room type
+    squareSideLength: square side length to compare with
+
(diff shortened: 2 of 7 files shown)


belgianguy:
To test the functionality, set a breakpoint in Temple.java where it passes its coordinateMatrix, then start the Singleplayer campaign mission with the white cilindrical tower region in front of the castle, you'll hit the breakpoint and get the coordinates of a 7x7 matrix, which at its center has a 5x5 square, so the hand is allowed to be added. (The intro cinematics also show this big temple).

src/toniarts/openkeeper/utils/RoomUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    public static Dimension maxArea(int[] hist) {
+
+        Dimension d = new Dimension();
+
+        final int n = hist.length;
+        final Deque<Integer> s = new ArrayDeque<Integer>();
+
+        int max_area = 0;
+        int tp;
+        int area_with_top;
```
ufdada: We usually use mixed case for variables instead of underscores (like areaWithTop). Basically we try to follow the regular Java naming conventions (http://www.oracle.com/technetwork/java/codeconventions-135099.html)

ufdada:
Another note: Seems like the hand in the original is always in the center of the water. In our implmentation it it always in the center of the room. The placement of the temple pillars is also wrong (they appear when there is no water next to them)

belgianguy:
In my local branch I've used a formula to calculate the "water" area of a temple based on its coordinatesMatrix, maybe this can then also be relied upon for the hand placement.

There is also a TileConstruction.CENTER_POOL variable that maybe has a function related to the Temple (as it has a pool in its center? Or is this something else?)

To be able to test all buildings I mentioned on IRC it maybe is a good idea to generate maps procedurally for use in ModelViewer to test different room combinations, as the Rooms look not quite right there.

E.g. the simplest of cases: the map is 5 x 5 tiles, claimed land owned by the "singleplayer" player.

If we can achieve that, add a room type of which we want to see the combinations possible. 
For example the Temple. 
Now for each combination we can think of, expand the map by 5x5 tiles and render the combination there.
eg 1x1, 2x1, 1x2,..4x3, 4x4 with a gap, 5 x 5, etc..., to see how they render.

This could provide a quick way to check for regression errors.

belgianguy:
I'll pick up the requested changes asap.

src/toniarts/openkeeper/utils/RoomUtils.java
```diff
@@ -0,0 +1,100 @@
+package toniarts.openkeeper.utils;
```
belgianguy: done

src/toniarts/openkeeper/utils/RoomUtils.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * Created by wietse on 17/04/17.
+ */
+public class RoomUtils {
+
+    /*
```
belgianguy: done

src/toniarts/openkeeper/utils/RoomUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    unless you're on the first row, then the value just becomes 1.
+    If you read a false, that means you're looking at a tile that is not part of the type, the histogram value gets reset to 0 (zero).
+
+    Then, for each row (histogram), calculate the max rectangle that can be fitted in the part that is TRUE with the maxArea function
+    if for any rectangle holds that its height and width equal or surpass squareSideLength, that means a square of
+    squareSideLength x squareSideLength fits in the TRUE part of the coordinates matrix.
+
+    Example: for the Temple we need at least a 5 x 5 square before the hand should be added.
+     */
+    public static boolean matrixContainsASquareWithSideLength(boolean [][] coordinatesMatrix, int squareSideLength) {
```
belgianguy: done

src/toniarts/openkeeper/utils/RoomUtils.java
```diff
@@ -0,0 +1,100 @@
...
+                return true;
+            }
+        }
+        return false;
+
+    }
+
+    public static Dimension maxArea(int[] hist) {
+
+        Dimension d = new Dimension();
```
belgianguy: done

src/toniarts/openkeeper/utils/RoomUtils.java
belgianguy: done

src/toniarts/openkeeper/utils/Dimension.java
```diff
@@ -0,0 +1,39 @@
+package toniarts.openkeeper.utils;
```
belgianguy: done

ufdada:
Seems like you are calculating the point **relative** to the temple room, but place it **absolute** to the map. You need to add also the position of the room. So instead of having a point 3/3 where the temple starts at 20/20 you need the center at 23/23.

belgianguy:
I've also added some WIP in my commit, to place the water (see commented code), so far no luck, I don't know how to build water tiles from Temple tiles (do I need access to the Terrain and set it to have Water flags?)...

tonihele:
toniarts.openkeeper.tools.convert.map.GameMap, it has water Terrain. You can get it from the KwdFile. Then just create an entityinstance (coordinates).

tonihele:
getWorldState().getLevelData().getMap().getWater() (getLevelData is deprecated, you can get it through GameState also..)

tonihele:
Do you want to try to implement it? It is really close :)

belgianguy:
I think I need to change the terrain flags. the Entity Instances I passed have Temple terrain as name, but the map water tiles have Water as name. I think we need some sort of hybrid solution, contains both Temple and Water properties.

tonihele:
> I think I need to change the terrain flags.

No? Terrain flags tell what the terrain is -> either temple or water. And water terrain actually draws only the water bed. The Water constructor just draws the water surface to wanted coordinates (creates a single plane, different than our normal batching).

tonihele:
`        EntityInstance<Terrain> ent = new EntityInstance<>(getWorldState().getLevelData().getMap().getWater());
        List<EntityInstance<Terrain>> instances = new ArrayList<>(1);
        instances.add(ent);
        // Add the coordinates to the ent!!!!
        Water.construct(assetManager, instances);`

belgianguy:
Ah, I see, I'll do that 👍

src/toniarts/openkeeper/world/room/Temple.java
```diff
@@ -102,19 +102,18 @@ protected BatchNode constructFloor() {
...
 
-//            if(waterArea[i][j]) {
-//                Terrain terrainAtPoint = roomInstance.getTerrainAtPoint(p);
-//                terrainAtPoint.getFlags().add(Terrain.TerrainFlag.WATER);
-//                EntityInstance<Terrain> entityInstance = new EntityInstance<>(terrainAtPoint);
-//
-//                waterBatches.add(entityInstance);
-//            }
+            if(waterArea[i][j]) {
+                EntityInstance<Terrain> ent = new EntityInstance<>(getWorldState().getGameState().getLevelData().getMap().getWater());
```
tonihele: This is a bit wrong idea-wise, EntityInstance is like a one continuous area of <T>. I don't remember anymore and I'm too lazy to look at the code at work. But normally we batch the map, 8x8 batches, rooms are batched separately. So is water areas (the surface). Water is batched a bit differently, it is homemade formula, we just create it as one mesh, which is quicker anyway. If you supply the water constructor with many many instances, it might batch them separately (here producing 1 tile patches, which make no sense), or it might work, I don't remember, but ideawise it is a bit wrong.

Batching is essential to performance, something that some of us dislike, but otherwise the object count in the scene would skyrocket. And would be very slow, even though the vertice count is small. Batching unites meshes with the same material to one single mesh.

So here, you would only have one instance, you can even init the intances arraylist capacity to 1. Create the entity. In the loop you would just add the coordinates to the one entity to you have. Thus you would end up with one water entity that spans an area. That is also less objects created and less functions calls :) Less carbage to collect and superior performance :)

tonihele: We have had unbatched version of the map rendering at the beginning, my GeForce 970 was brought to its knees when we had all the objects and the map. We didn't even have creatures back then. The ye olden days.

tonihele:
Screenshots! :) Let's adjust the height :)

belgianguy:
@ufdada : could you explain more about the comment about the rendering of the pillars next to water?

belgianguy:
On second though the offset seems wrng, not the height, or I'm going crazy.

closer

tonihele:
Hmm yes, it seems not to be on the right spot. If you start with -debug flag, you can at least see the coordinates on screen and compare them to the ones you add there.

tonihele:
```
            int i = count / waterArea[0].length;
            int j = count % waterArea[0].length;

            if(waterArea[i][j]) {
                ent.addCoordinate(p);
            }

count++;
```
^ this feels kinda fishy. I don't understand this, could it be that here is something going on?

tonihele:
I didn't test the code, but I think you are creating the matrix, it kinda uses a fake matrix. A local minimal matrix. Everything else is in local coordinates then.

ufdada:
@belgianguy 
temple

Only the one with the water should have pillars, the other one originally have temple candlesticks in the corners that point outside.

belgianguy:
@tonihele : you were spot on, it was the tranformation from world to local what I was attempting.

Now I also Now also added water to corner tiles (NE || SE || SW || NW) to make the body of water seamless (otherwhise the water did not meet the rounded corner sections)  of the temple borders.

But I don't know if e.g. a Vampire would now take damage if he's on such a border tile, or if it won't affect him as water is rendered a little lower than the floor he's on (temple border).

belgianguy:
no_center
only_center
sticks

ufdada:
Looks like these candlesticks are always on temple1 tile (the corner tile) if there are 3 other tiles types arround it (pic 2 has no candleholder pointing inward).

Funny enough even dk2 doesn't update correctly:
shot0016
In the second picture all i did was removing the tile and put it on again

belgianguy:
big_temple
temple2

belgianguy:
Trying to think of some sort of logic:
- Using the coordinates matrix.
- Subdivide this rectangle in 4 equal rectangles (rows / 2, cols / 2) so you become an upperLeft, upperRight, bottomLeft and bottomRight square.
- Then, for each tile that is not part of the roomType (FALSE), check in what sub-square it is. Store that type somewhere.
- Check if it has any direct neighbours (not diagonal) that are a "Border" tile. (Breder == RoomType but at least has one non-Roomtype tile as neighbour (diagonal included) and that any of the Border tile's  direct (not diagonal) neighbour's don't already have candles.. 

If so, depending on your subsquare and where the neigbour is located, add a candle to that tile.
UpperLeft SubSquare + tile to the right = add candle to the tile in the topLeft corner
UpperLeft SubSquare + tile above = add candle to the tile in the bottomLeft corner
UpperLeft SubSquare + tile below = add candle to the tile in the topLeft corner
UpperLeft SubSquare + tile to the left = add candle to the tile in the topRight corner

UpperRight + tile to the right = add candle to the tile in the topLeft corner
UpperRight + tile above = add candle to the tile in the bottomRight corner
UpperRight + tile below = add candle to the tile in the topRight corner
UpperRight + tile to the left = add candle to the tile in the topRight corner

etc...

(Egde case "no non-roomtype" in coordinates, room is square, take topLeft, bottomRight and add the candles to each corner)

tonihele:
> Funny enough even dk2 doesn't update correctly:

Yeah, it is with every room. You can build a giant hatchery for example with just few coops. DK II doesn't reconstruct the room entirely. This is something that I didn't plan to honor, as I don't even know the logic fully on this partial update. And it seems a bit unnecessary, IMHO doesn't really add anything to the game. Might even break something since you get less room objects depending the way you build the room.

tonihele:
For the candle sticks.... Seems awfully a lot of Workshop... @belgianguy try copying the pillars from the Workshop.java. See if it works straight...

tonihele:
> But I don't know if e.g. a Vampire would now take damage if he's on such a border tile, or if it won't affect him as water is rendered a little lower than the floor he's on (temple border). Visually we'd best keep the water there, but maybe should not let it do damage there. (TileConstruction.CENTER_POOL maybe? :p)

We don't have physics or anything of that sort yet. But I think these are rather simple, we just check if the character is below the sea level. The see the tile which it is on, if WATER, then creature in water. Temple would not trigger this at any time. Even with real collisions, I don't think the water surface would trigger this.

We also have effects that require this knowledge, maybe too costly for real physics check for every particle.

belgianguy:
Okay, so the candles are solved for now (they don't have a falme yet but that is known).

TEMPLE10 and TEMPLE12 are the pieces which contain arches, so I'd guess these only should be used when they're next to water, and otherwise use another quarter tile instead.

However, this logic is in DoubleQuad's constructQuad, which is very generic, and we'd need a very specific change for the Temple, anyone any pointers on how to approach this?

ufdada:
I think in this case we probably have to make a separate implementation of the double quad since i don't see any similarities right now. If we find one we can improve it later on.

Looks good so far!

The only (very minor) thing i noticed if you make a room like shown in the comment it still shows candles (the right half is the correct candle positioning).
https://github.com/tonihele/OpenKeeper/pull/279#issuecomment-296175766

This is how it looks right now (the candle in the circles shouldn't be there):
main1

tonihele:
It would be nice to showcase this as well :) I'm planning to do a progress video VERY soon. It has been so long since we are least heard of.

belgianguy:
schermafdruk van 2017-05-11 21-25-34
schermafdruk van 2017-05-11 21-26-04
schermafdruk van 2017-05-11 21-27-58
schermafdruk van 2017-05-11 21-28-15

belgianguy:
Okay, so far no more pillars when there is no water, but the other tiles (piece nr 11 is an arch corner) are misplaced as well. Checking combinations by hand and editing them in the debugger.

Tiles contain 4 small pieces, the render NW, NE, SW, SE, for the groups of 4, it will work row by row, only starting the next row if the current one is completed.

1 is a corner piece with 2 borders (needs orienting in the right direction)
0 is a side piece with 1 border (needs orienting in the right direction)
2 is a piece without borders and is used to 'fill' inside

Given 1 and 0 are oriented right, a 2 x 4 tiles would look like this:

10 01
02 20
02 20
10 01

Still need to flush out some edge cases wrt border rotating and some 2 tiles that show up that should be 0.

testsrc/toniarts/openkeeper/world/room/TempleTest.java
```diff
@@ -0,0 +1,218 @@
...
+        expected.add("Temple0");
+        expected.add("Temple2");
+        expected.add("Temple1");
+        expected.add("Temple0");
+
+        final List<String> actual = captor.getAllValues();
+
+        final List<Float> rotationAnglesActual = captorFloat.getAllValues();
+        final List<Float> rotationAnglesExpected = new ArrayList<>(4);
+        rotationAnglesExpected.add(1.5707964f);
```
ArchDemons: Use FastMath.PI or .HALF_PI, not your constants

testsrc/toniarts/openkeeper/world/room/TempleTest.java
```diff
@@ -0,0 +1,218 @@
...
+        rotationAnglesExpected.add(-1.5707964f);
+        rotationAnglesExpected.add(3.1415927f);
+        rotationAnglesExpected.add(3.1415927f);
+
+        final List<Vector3f> vectorsActual = captorVector.getAllValues();
+        final List<Vector3f> vectorsExpected = new ArrayList<>(4);
+        vectorsExpected.add(new Vector3f(-0.25f,0,-0.25f));
```
ArchDemons: -0.25f and 0.25f must be scaled by SCALE_WIDTH constant.

testsrc/toniarts/openkeeper/world/room/TempleTest.java
belgianguy: Done. :)

testsrc/toniarts/openkeeper/world/room/TempleTest.java
belgianguy: Done! :)

tonihele:
Notify when ready & conflicts resolved :)

belgianguy:
Getting closer for sure, but still errors remain:
* My check of a tile being a border tile is too broad, I'm afraid it'll have to be done at the "piece" level. 
Now the (water) pillars also show up at the NW point of a tile if it has water to its SE
* Odd combinations still have a weird behaviour from time to time, but a lot of them have been solved as well.

tonihele:
Thank you @belgianguy for you work. Sorry it didn't make it. I'll close this now as this issue is now resolved with other code.

Of course feel free to check the current code and optimize or add the tests to that etc.

belgianguy:
No problem! :) Progress was made, and that's what ultimately counts!

PR #280, opened 2017-04-17
Clean asset folders before converting

ufdada:
This clean the folders where the assets are stored before a new conversion is happening. This should avoid having the need to manually remove the assets.

Since we don't really respect custom models / textures (because they may be overwritten in the conversion process anyway) yet i see it less problematic right now.

### DIFF
--- a/src/toniarts/openkeeper/tools/convert/AssetsConverter.java
+++ b/src/toniarts/openkeeper/tools/convert/AssetsConverter.java
@@ -100,7 +100,7 @@ public abstract class AssetsConverter {
         PATHS(4),
         HI_SCORES(2),
         FONTS(3),
-        MAP_THUMBNAILS(2);
+        MAP_THUMBNAILS(3);
 
         private ConvertProcess(int version) {
             this.version = version;
--- a/src/toniarts/openkeeper/tools/convert/AssetsConverter.java
+++ b/src/toniarts/openkeeper/tools/convert/AssetsConverter.java
@@ -145,7 +145,7 @@ public abstract class AssetsConverter {
     public static final String TEXTURES_FOLDER = "Textures";
     public static final String SPRITES_FOLDER = "Sprites";
     public static final String MAP_THUMBNAILS_FOLDER = "Thumbnails";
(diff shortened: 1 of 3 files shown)


ArchDemons:
Please. Add one more commit. Need to change Sprites folder in AssetsConverter.java

`File destinationFolder = new File(getAssetsFolder().concat(TEXTURES_FOLDER).concat(File.separator).concat("Sprites/"));`

change to 

`File destinationFolder = new File(getAssetsFolder().concat(TEXTURES_FOLDER).concat(File.separator));`

ufdada:
Good find, i forgot about those. But i want to do a different approach so the texture folder doesn't get to cluttered like it already is. Also they would have been deleted if the texture version changed.

src/toniarts/openkeeper/tools/convert/AssetsConverter.java
```diff
@@ -142,11 +143,12 @@ public String toString() {
     public static final String MATERIALS_FOLDER = "Materials";
     public static final String MODELS_FOLDER = "Models";
     public static final String TEXTURES_FOLDER = "Textures";
-    public static final String MAP_THUMBNAILS_FOLDER = TEXTURES_FOLDER + File.separator + "Thumbnails";
-    public static final String MOUSE_CURSORS_FOLDER = "Interface" + File.separator + "Cursors";
-    public static final String FONTS_FOLDER = "Interface" + File.separator + "Fonts";
-    public static final String TEXTS_FOLDER = "Interface" + File.separator + "Texts";
-    public static final String PATHS_FOLDER = "Interface" + File.separator + "Paths";
+    public static final String MAP_THUMBNAILS_FOLDER = "Thumbnails";
+    public static final String INTERFACE_FOLDER = "Interface" + File.separator;
```
tonihele: A small thing, but since this folder holds no significance (?, only used internally), change to private modifier? :) Very small thing...

tonihele:
This makes a lot of sense now since the files in Converted shouldn't be tampered with. We have the mirror where any modifications go.

ufdada:
interface folder is now private :)

PR #281, opened 2017-04-19
Add commands to the console for easier debugging

ufdada:
Since the console was still unused i gave it a shot and implemented some commands for it.

Debug flag + F11 is still needed.

### DIFF
--- a/assets/Interface/ConsoleStyle.xml
+++ b/assets/Interface/ConsoleStyle.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<nifty-styles>
+    <!-- console background -->
+    <style id="nifty-console-panel">
+        <attributes childLayout="vertical" backgroundColor="#333b"/>
+    </style>
+
+    <!-- console line -->
+    <style id="nifty-console-line">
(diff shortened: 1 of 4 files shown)


ufdada:
Seems like the setFocus on the textfield sometimes interferes with the listbox content (it disappears at first, but not always).

I already tried to clear the console (because if i do that the scrolling in the console history works again). Also the effects (onShow, onHide) don't seem to have an impact on this. But the only thing really helped was to not set the focus on the textfield at setting it visible.

Any ideas on this?

tonihele:
Not entirely what is the problem you described? Seems to work nicely. Ok to merge if your question doesn't mean you are still trying to fix something. Professional looking console we now have :)

ufdada:
> Not entirely what is the problem you described? Seems to work nicely.

I'm going nuts, because yesterday _sometimes_ it failed to display the message (with the commands), you had to have another command typed in and then it showed it. Then the listbox scrolling wouldn't work (you could move the scrollbar and the text wouldn't move). Only clearing the console would fix it.

Can't reproduce it now, just tried it 12 times (yesterday it was at least at the second attempt). I blame it on the jme debug mode. So i kinda give up on this as it wasn't introduced with these changes anyway.

PR #283, opened 2017-04-22
Debriefing and Game HUD

ArchDemons:
added GameHUD objective information
added toggle bottom panel (need redesign)
added debriefing screen (not tested)
added zoom to dungeon from GameHUD
fixed some GameHUD tooltip text (with placeholders)
fixed seldom NPE

### DIFF
--- a/src/toniarts/openkeeper/game/state/MainMenuScreenController.java
+++ b/src/toniarts/openkeeper/game/state/MainMenuScreenController.java
@@ -49,6 +49,7 @@ import java.util.List;
 import java.util.logging.Logger;
 import toniarts.openkeeper.Main;
 import toniarts.openkeeper.game.data.CustomMPDLevel;
+import toniarts.openkeeper.game.data.GameResult;
 import toniarts.openkeeper.game.data.HiScores;
 import toniarts.openkeeper.game.data.Keeper;
 import toniarts.openkeeper.game.data.Level;
@@ -69,6 +70,7 @@ public class MainMenuScreenController implements IMainMenuScreenController {
 
     public final static String SCREEN_EMPTY_ID = "empty";
     public final static String SCREEN_START_ID = "start";
+    public final static String SCREEN_DEBRIEFING_ID = "debriefing";
 
     private final MainMenuState state;
     private Nifty nifty;
@@ -787,7 +789,7 @@ public class MainMenuScreenController implements IMainMenuScreenController {
             mainObjectiveImage.setHeight(img.getHeight());
             mainObjectiveImage.show();
         } catch (Exception e) {
-            logger.warning("Can't find image " + objectiveImage.replace("$index", "1"));
+            logger.warning("Can't find image " + objectiveImage.replace("$index", "0"));
             mainObjectiveImage.hide();
         }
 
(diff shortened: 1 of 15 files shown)


src/toniarts/openkeeper/game/state/MainMenuScreenController.java
```diff
@@ -828,4 +830,76 @@ private void showBriefing() {
...
+        }
+
+        boolean levelWon = result.getData(GameResult.ResultType.LEVEL_WON);
+        screen.findNiftyControl("levelWon", Label.class).setText(levelWon ? "${menu.21}" : "${menu.22}");
+        int timeTaken = (int) result.getData(GameResult.ResultType.TIME_TAKEN);
+        screen.findNiftyControl("timeTaken", Label.class).setText(timeToString(timeTaken));
+    }
+
+    private String timeToString(int time) {
+        String result = "";
```
ufdada: Wouldn't DateTimeFormatter do the job as well? At least this always seems more straight forward to me instead of writing the an own formatter.

ArchDemons: I am not sure that it will give the same result as my function (outputs like in the original game). If DateTimeFormatter not output days if days = 0, hours - if 0, minutes - if 0, than I can use it.

tonihele: An age old problem :( Java 8 got those long waited improvements to Time/Date, but I'm not too sure that there is a formatter that does exactly this. Probably still need to construct the format string with these IFs and ELSEs to use in the formatter. Perhaps using Duration might look better but a small thing.

I'm not too familiar with the new Java 8 time/date stuff. Worth to study though.

ufdada: I'm sorry, i wasn't looking correct, this isn't a date at all :).

But it seem like the always use at least 2 digits for the playtime. How about `String.format(":%02d", seconds)` ?

tonihele: Or those TimeUnit.toXXXX instead of math.

src/toniarts/openkeeper/game/state/MainMenuScreenController.java
```diff
@@ -828,4 +830,76 @@ private void showBriefing() {
             }
         }
     }
+
+    public void showDebriefing(GameResult result) {
+        goToScreen(SCREEN_DEBRIEFING_ID);
+
+        Label levelTitle = screen.findNiftyControl("dLevelTitle", Label.class);
```
ufdada: I get here a NPE. Seems like the screen isn't switched right there. ScreenId says "empty".

Maybe just using `nifty.getScreen(SCREEN_DEBRIEFING_ID)` ?

src/toniarts/openkeeper/game/state/MainMenuScreenController.java
```diff
@@ -828,4 +830,76 @@ private void showBriefing() {
...
+                subObjectiveImage.show();
+            } catch (Exception e) {
+                logger.warning("Can't find image " + objectiveImage.replace("$index", "1"));
+                subObjectiveImage.hide();
+            }
+        }
+
+        boolean levelWon = result.getData(GameResult.ResultType.LEVEL_WON);
+        screen.findNiftyControl("levelWon", Label.class).setText(levelWon ? "${menu.21}" : "${menu.22}");
+        int timeTaken = (int) result.getData(GameResult.ResultType.TIME_TAKEN);
```
ufdada: This gives a NumberFormatException because it is a float value. Maybe use `Math.round`?

ArchDemons:
win level cheat the best thing to test debriefing screen. I am sick of going through the first level again

ufdada:
Since it needs some work to function properly you could add at Line 212 in Playerinteractionstate:
```
case WIN_LEVEL:
    gameState.setEnd(true);
    break;
```

You still need to press Space

ArchDemons:
wait sound PR because sound assets reconverted in new folders. MainMenuState does not work

tonihele:
> win level cheat the best thing to test debriefing screen. I am sick of going through the first level again

We'll have save games "soon" :) But you can also craft a level1 with almost instant win.

PR #284, opened 2017-04-22
Sound

ArchDemons:
Test new sound system ^_^.  Need some magic with gui sounds. But not in this PR

### DIFF
--- a/src/toniarts/openkeeper/game/sound/SoundCategory.java
+++ b/src/toniarts/openkeeper/game/sound/SoundCategory.java
@@ -33,21 +33,26 @@ import toniarts.openkeeper.utils.PathUtils;
  * @author archdemon
  */
 public class SoundCategory {
+    // default sounds categories
+    public static final String SPEECH_MENTOR = "SPEECH_MENTOR";
+    public static final String SPEECH_MULTIPLAYER = "SPEECH_MULTIPLAYER";
+    public static final String SPEECH_HORNY = "SPEECH_HORNY";
+    // TODO need to add some GUI sounds categories
     private final String folder;
     private final String name;
     private final Map<Integer, SoundGroup> groups = new HashMap<>();
 
     private static final Logger LOGGER = Logger.getLogger(SoundCategory.class.getName());
 
-    public SoundCategory(String folder, String name) {
+    public SoundCategory(String name, boolean useGlobal) {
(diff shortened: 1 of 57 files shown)


src/toniarts/openkeeper/game/sound/SoundCategory.java
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Get SdtFile
+     *
+     * @param archiveFilename name in BankMapFile
+     * @return SdtFile
+     */
+    @Nullable
+    public static SdtFile getSdtFile(String archiveFilename) {
+
+        for (String part : new String[] {"HD.sdt", "HD.SDT", "HW.sdt", "HW.SDT"}) {
```
ufdada: Shouldn't `ConversionUtils.getRealFileName` be better instead of duplicate entries?

tonihele: I never understood the difference between HD & HW... Now we prefer HD files, is that just a guess?

tonihele:
Oh nice, you can solve conflicts in Github straight...

tonihele:
Hmm, there is a lot of great work there, but what is the rationale on extracting the HDs and HWs? There are now 2000 duplicate sounds. Not really an issue on today's super computers, but doesn't feel so elegant? Maybe better for debugging.

ArchDemons:
when we decide what better HD or HW, we can delete unnecessary files

tonihele:
But changing it with this kind of system requires some kind of redoing? Or did it already have a switch somewhere? I'm just afraid the filenames get more hard coded with the path.

ArchDemons:
It will not be a problem to redo it. Conversely, now HW and HD create more problems

ArchDemons:
I heard the difference. HD - clear sound, HW  - as with echo (the environment) in an empty room or in a dungeon.

tonihele:
Hmm, maybe there is a flag or something to use environmental or pure. Was it that there was more HWs, or which way around it was?

If we find no logic which to play. We could use HD always, and add an option for environmental sound which would mean those OpenAL environmental audio, might be better quality than pre-made low quality files although of course not the same.  And this is just an assumption, I have no idea about sound. And I listen to way too much heavy metal.

PR #285, opened 2017-04-25
Additional commands

ufdada:
I've added several new commands and also implemented "level up" and "win level" cheats.


tonihele:
Nice, I hope that level winning helps that Horny feature creation.

PR #286, opened 2017-05-01
Fix creature cards

ufdada:
Not all changes from
https://github.com/tonihele/OpenKeeper/commit/73983d6e8d316c17942c2695aed7083e285963bc
where restored in
https://github.com/tonihele/OpenKeeper/commit/775fd294f31a574eac65a213ef3eb95e0840bd2f

Even though it never crashed, it showed several NPEs in the console log since the ids of some elements weren't correct, leaving the creature tab complely useless (ALL creatures had their own card and thus couldn't be picked up and the imp count was always zero).

Also rearranged the secondary click for the imps, since nifty seem to dislike if its defined before the primary.


tonihele:
Great, thanks!

PR #294, opened 2017-05-21
Movie unlock

ufdada:
This implements the missing check for the movie list. Movies can now be unlocked via winning a campaign level or by adding `LEVEL_STATUS7(string)=COMPLETED` in the user properties (only if you are a cheater 😉)


tonihele:
Cool :) Didn't try it but looks ok.

PR #344, opened 2019-10-14
save opaque textures as RGB pngs

Trass3r:
saves space and also makes them usable in DK2


tonihele:
Cool!

tonihele:
3,1MB saved. 0,34% improvement. So maybe not bump the textures version (triggers conversion). But anyway it is great!

PR #345, opened 2019-10-15
Upgrading Gradle Wrapper from 4.10.3 to 5.6.2

Thunderforge:
On my machine, this upgrade makes `gradlew clean build` about 25% faster. It also opens the door to building the project on Java 9+, should we choose to upgrade some day.


tonihele:
Update to Java 11 is planned. Right after jMonkeyEngine SDK supports it. For sentimental reasons, I'm attached to it. It shouldn't be that far anymore. I've already tested a version that it based on new Netbeans, thus providing this support.

Thunderforge:
My reference to Java 9+ was that it's only supported by newer versions of Gradle. So by upgrading to the latest Gradle, that barrier is removed.

tonihele:
Yeah, no problems there. Thanks for this.

PR #364, opened 2020-04-12
Decay job fix

ArchDemons:
first COVID-19 pull request

### DIFF
--- a/src/toniarts/openkeeper/game/controller/MapController.java
+++ b/src/toniarts/openkeeper/game/controller/MapController.java
@@ -285,8 +285,8 @@ public final class MapController extends Container implements Savable, IMapContr
     }
 
     @Override
-    public boolean isSelected(int x, int y, short playerId) {
-        MapTile tile = getMapData().getTile(x, y);
+    public boolean isSelected(Point p, short playerId) {
+        MapTile tile = getMapData().getTile(p);
         if (tile == null) {
             return false;
         }
@@ -294,8 +294,8 @@ public final class MapController extends Container implements Savable, IMapContr
     }
 
     @Override
-    public boolean isTaggable(int x, int y) {
-        MapTile tile = getMapData().getTile(x, y);
+    public boolean isTaggable(Point p) {
+        MapTile tile = getMapData().getTile(p);
         if (tile == null) {
             return false;
         }
@@ -304,8 +304,8 @@ public final class MapController extends Container implements Savable, IMapContr
     }
 
     @Override
-    public boolean isBuildable(int x, int y, short playerId, short roomId) {
-        MapTile tile = getMapData().getTile(x, y);
+    public boolean isBuildable(Point p, short playerId, short roomId) {
+        MapTile tile = getMapData().getTile(p);
         if (tile == null) {
             return false;
         }
@@ -355,8 +355,8 @@ public final class MapController extends Container implements Savable, IMapContr
     }
 
     @Override
-    public boolean isClaimable(int x, int y, short playerId) {
-        MapTile tile = getMapData().getTile(x, y);
+    public boolean isClaimable(Point p, short playerId) {
+        MapTile tile = getMapData().getTile(p);
         if (tile == null) {
             return false;
         }
@@ -385,9 +385,8 @@ public final class MapController extends Container implements Savable, IMapContr
     }
 
     @Override
-    public boolean isSellable(int x, int y, short playerId) {
-        MapTile tile = getMapData().getTile(x, y);
-        Point p = new Point(x, y);
+    public boolean isSellable(Point p, short playerId) {
+        MapTile tile = getMapData().getTile(p);
         if (tile.getOwnerId() == playerId && getRoomCoordinates().containsKey(p)) {
 
             // We own it, see if sellable
@@ -398,8 +397,8 @@ public final class MapController extends Container implements Savable, IMapContr
     }
 
     @Override
-    public boolean isWater(int x, int y) {
-        MapTile tile = getMapData().getTile(x, y);
+    public boolean isWater(Point p) {
+        MapTile tile = getMapData().getTile(p);
         if (tile == null) {
             return false;
         }
@@ -408,8 +407,8 @@ public final class MapController extends Container implements Savable, IMapContr
     }
 
     @Override
-    public boolean isLava(int x, int y) {
-        MapTile tile = getMapData().getTile(x, y);
+    public boolean isLava(Point p) {
+        MapTile tile = getMapData().getTile(p);
         if (tile == null) {
             return false;
         }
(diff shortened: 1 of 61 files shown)


src/toniarts/openkeeper/game/state/loading/LoadingState.java
```diff
@@ -74,6 +79,9 @@ public void initialize(final AppStateManager stateManager, final Application app
 
         // Load up the title screen
         setupTitleScreen();
+
+        loadingThread = new LoadingThread(this.name);
+        loadingThread.start();
```
tonihele, line 84: The problem with this.... While it is quite ugly to just create a new thread when you create this, but the loading will start immediately. Now instead the loading is governed by the framerate. Well, no really that big of an issue. But what if you load the app minimized or alt tab away. Loading basically never starts because it waits for the screen to draw.

src/toniarts/openkeeper/game/controller/MapController.java
```diff
@@ -285,17 +285,17 @@ public void selectTiles(Vector2f start, Vector2f end, boolean select, short play
     }
 
     @Override
-    public boolean isSelected(int x, int y, short playerId) {
-        MapTile tile = getMapData().getTile(x, y);
+    public boolean isSelected(Point p, short playerId) {
```
tonihele, line 288: Overload would have been better. Now we always need to construct a point. I guess the performance and memory hit is quite marginal, even I at some point wanted to have a point cache. Waiting for tuples to come to Java as well. Then we wouldn't need to construct these objects.

src/toniarts/openkeeper/game/state/GameState.java
```diff
@@ -323,6 +322,14 @@ public void onLoadComplete() {
                 GameState.this.stateManager.getState(PlayerState.class).setEnabled(true);
                 GameState.this.stateManager.getState(SoundState.class).setEnabled(true);
 
+                // FIXME PlayerCameraState should initialized before we start
+                while (true) {
```
tonihele, line 326: Hmm, got scared there for a moment :) This class is not used.

src/toniarts/openkeeper/game/state/loading/LoadingState.java
ArchDemons: This is not a bug, this is feature ;-). I can combine our code. Pass name and app to the constructor. Combine?

src/toniarts/openkeeper/game/task/worker/DigTileTask.java
```diff
@@ -87,13 +87,14 @@ public String toString() {
 
     @Override
     public void executeTask(ICreatureController creature, float executionDuration) {
-
+//        if (creature.getMeleeAttack().execute()) {
```
tonihele, line 90: Hmm?

src/toniarts/openkeeper/game/controller/MapController.java
ArchDemons: We must use the Point everywhere in methods. It is better to load from files to a Point and getX() methods, getY() to do through a Point.

src/toniarts/openkeeper/world/creature/steering/AbstractCreatureSteeringControl.java
```diff
@@ -47,8 +47,8 @@
     private boolean tagged;
     private boolean independentFacing = false;
     private float maxLinearSpeed = 1;
-    private float maxLinearAcceleration = 2;
-    private float maxAngularSpeed = 10.0f;
+    private float maxLinearAcceleration = 4;
+    private float maxAngularSpeed = 20.0f;
```
tonihele, line 51: How is the steering now?

src/toniarts/openkeeper/game/state/GameState.java
ArchDemons: In a far, far-away galaxy, the absence of this code threw an exception. I corrected in the old code so as not to forget why errors occur

src/toniarts/openkeeper/game/state/loading/LoadingState.java
tonihele: Yeah the other changes I think are just fine. I would be conservative about the thread starting yes. So yes :)

src/toniarts/openkeeper/game/controller/MapController.java
tonihele: Ok

src/toniarts/openkeeper/world/creature/steering/AbstractCreatureSteeringControl.java
ArchDemons: When this code worked, it was better. Now it needs to be transferred to another class. I don’t know which one. I wrote down here so as not to forget

src/toniarts/openkeeper/game/task/worker/DigTileTask.java
ArchDemons: When we execute the task, we must check whether it is possible to complete it. Specifically for DigTile, you need to check if the creature’s meele attack is recharged.

src/toniarts/openkeeper/world/creature/steering/AbstractCreatureSteeringControl.java
tonihele: Ah, oh yeah, true. I thought this was the class we used. Yep, this is deprecated as well. The new one is in toniarts.openkeeper.game.navigation.steering.*

src/toniarts/openkeeper/game/task/worker/DigTileTask.java
tonihele: Yeah, that needs a bit discussion that how we generally do that. Now I have defaulted all behavior to _measurement / per second_. The other way I guess would the to sync everything with the animations, that is what you are after here? I don't really have a clear preference which one is better and what the original game used.

ArchDemons:
all done. @tonihele, you can see new steering. Not perfect, but funny :). This is not final changes of steering

src/toniarts/openkeeper/game/task/worker/DigTileTask.java
ArchDemons: DigTile recharge == Melee attack recharge. Claim tile recharge == Some imp spell (or other attack) recharge

ArchDemons:
I did the room construction as in the original game (with delay). But the code worked only in the old version. I need names of new classes so that I can port the code there.

tonihele:
> 
> 
> all done. @tonihele, you can see new steering. Not perfect, but funny :). This is not final changes of steering

Thanks! Well, as long as it is not too funny and still functional. We are trying to keep this in a playable state always in the master. Any non-ready stuff should be kept in their respective branches.

ArchDemons:
> > all done. @tonihele, you can see new steering. Not perfect, but funny :). This is not final changes of steering
> 
> Thanks! Well, as long as it is not too funny and still functional. We are trying to keep this in a playable state always in the master. Any non-ready stuff should be kept in their respective branches.

Now they do not pass through the walls. It's good. But I still don't like the way they walk. And I don’t know how to change this

tonihele:
Yeah, they look like they are a bit drunk now but it seems to work.

tonihele:
Once approved it can be merged by the author. Sounds logical workflow :)

> I did the room construction as in the original game (with delay). But the code worked only in the old version. I need names of new classes so that I can port the code there.

toniarts.openkeeper.game.controller.GameWorldController

PR #368, opened 2020-04-13
Feature 155

ArchDemons:
Need some help. Method `isSellable(SelectionArea ...` in class `PlayerInteractionState` (method `getColorIndicator()`) not working. Problem with access in other thread. Property `roomCoordinates` is empty in class `MapController`, but should not

### DIFF
--- a/src/toniarts/openkeeper/game/controller/GameWorldController.java
+++ b/src/toniarts/openkeeper/game/controller/GameWorldController.java
@@ -305,11 +305,9 @@ public class GameWorldController implements IGameWorldController, IPlayerActions
 
     @Override
     public void build(SelectionArea area, short playerId, short roomId) {
-//        int x1 = (int) Math.max(0, start.x);
-//        int x2 = (int) Math.min(kwdFile.getMap().getWidth(), end.x + 1);
-//        int y1 = (int) Math.max(0, start.y);
-//        int y2 = (int) Math.min(kwdFile.getMap().getHeight(), end.y + 1);
-        mapController.buildOrSellRoom(area, playerId, roomId);
+        if (mapController.isBuildable(area, playerId, roomId)) {
+            mapController.buildOrSellRoom(area, playerId, roomId);
+        }
     }
 
     @Override
@@ -332,11 +330,6 @@ public class GameWorldController implements IGameWorldController, IPlayerActions
 
         // If this is a bridge, we only got the starting point(s) as valid so we need to determine valid bridge pieces by our ourselves
         Room room = kwdFile.getRoomById(roomId);
-//        if ((room.getFlags().contains(Room.RoomFlag.PLACEABLE_ON_WATER))
-//                || room.getFlags().contains(Room.RoomFlag.PLACEABLE_ON_LAVA)) {
-//            instancePlots = new ArrayList(mapController.getTerrainBatches(instancePlots, x1, x2, y1, y2));
-//        }
-
         // See that can we afford the building
         synchronized (GOLD_LOCK) {
             int cost = instancePlots.size() * room.getCost();
@@ -428,7 +421,10 @@ public class GameWorldController implements IGameWorldController, IPlayerActions
 
     @Override
     public void sell(SelectionArea area, short playerId) {
-        mapController.buildOrSellRoom(area, playerId);
+        // FIXME does not work isSellable
+//        if (mapController.isSellable(area, playerId)) {
+            mapController.buildOrSellRoom(area, playerId);
+//        }
     }
 
     @Override
--- a/src/toniarts/openkeeper/game/controller/GameWorldController.java
+++ b/src/toniarts/openkeeper/game/controller/GameWorldController.java
@@ -75,6 +75,7 @@ import toniarts.openkeeper.tools.convert.map.Terrain;
 import toniarts.openkeeper.tools.convert.map.Tile;
 import toniarts.openkeeper.tools.convert.map.Variable;
 import toniarts.openkeeper.utils.WorldUtils;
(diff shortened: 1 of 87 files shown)


src/toniarts/openkeeper/game/controller/GameWorldController.java
```diff
@@ -75,6 +75,7 @@
 import toniarts.openkeeper.tools.convert.map.Tile;
 import toniarts.openkeeper.tools.convert.map.Variable;
 import toniarts.openkeeper.utils.WorldUtils;
+import toniarts.openkeeper.view.selection.SelectionArea;
```
tonihele, line 78: We can't mix these like this. High hygiene levels must be maintained, especially during Covid season.

The logic needs to be separate from the view. If some common datatransfer etc. class is needed, it needs to be in another package. The VIEW basically uses logic classes yes, but definitely NOT the other way around.

src/toniarts/openkeeper/game/controller/GameWorldController.java
```diff
@@ -329,10 +332,10 @@ public void build(Vector2f start, Vector2f end, short playerId, short roomId) {
 
         // If this is a bridge, we only got the starting point(s) as valid so we need to determine valid bridge pieces by our ourselves
         Room room = kwdFile.getRoomById(roomId);
-        if ((room.getFlags().contains(Room.RoomFlag.PLACEABLE_ON_WATER))
-                || room.getFlags().contains(Room.RoomFlag.PLACEABLE_ON_LAVA)) {
-            instancePlots = new ArrayList(mapController.getTerrainBatches(instancePlots, x1, x2, y1, y2));
-        }
+//        if ((room.getFlags().contains(Room.RoomFlag.PLACEABLE_ON_WATER))
```
tonihele, line 335: Obsolete now? Shouldn't have code in comments, unless that is commented, otherwise very confusing later

src/toniarts/openkeeper/game/controller/GameWorldController.java
ArchDemons: Perhaps the initial location of the SelectionArea class is not very good. It does not contain any visual component. All visual part contains in `toniarts.openkeeper.view.selection.SelectionHandler`

src/toniarts/openkeeper/tools/convert/map/Thing.java
```diff
@@ -16,11 +16,13 @@
  */
 package toniarts.openkeeper.tools.convert.map;
 
+import com.jme3.network.serializing.serializers.EnumSerializer;
```
tonihele, line 19: Yeah.. I guess this is ok for now, since it was already misused. But jME or network classes don't belong here. Same hygiene stuff.

src/toniarts/openkeeper/game/controller/GameWorldController.java
tonihele: I have tossed these common classes to toniarts.openkeeper.common, but also toniarts.openkeeper.game.data might be a place for this. Dunno, but not view if this is going to be used in logic.

And if this goes through network, as slim as possible. All you can generate or calculate, can be done so. Cached with transient etc fields.

src/toniarts/openkeeper/game/controller/GameWorldController.java
ArchDemons: Must be deleted

tonihele:
Rooms are not yet transferred to the client. The reason being also that I was unsure what is the smartest way to do this. Currently it is just that raw map data as read only data.

Clients can parse these from the map data they have yes. But in addition to that, the clients need some information about the room usages and whatnots in the tooltip. Clients do also kinda have this information, but it needs a lot of processing. Like they know that when an entity is stored/using some room. Also they can calculate room capacities and all that....

One idea was that rooms are also entities. And the map has references to entityIDs. Server can preprocess the usage data to components. This would be really easy solution and allow the rooms to work fully via ES network structure.

That also brings the idea of map tiles being entities as well. Updating and all that would also then be like with the others.

src/toniarts/openkeeper/game/controller/GameWorldController.java
ArchDemons: SelectionArea contains 2 Vector2f fields. Transmission surrendered a bit more expensive

src/toniarts/openkeeper/tools/convert/map/Thing.java
ArchDemons: Agree

src/toniarts/openkeeper/game/controller/GameWorldController.java
tonihele: Well, if it is the same two fields that already went there, that is ok. This is from user interaction (meaning not all the time). So the extra layer is just few bytes.

ArchDemons:
o prevent possible cheats, we should not transmit all data about other players, if the current player does not see other players. In this regard, each client machine should not contain relevant information about rooms and tiles. Only Server should have all information

src/toniarts/openkeeper/game/controller/GameWorldController.java
ArchDemons: I generally think we need to use the Point class as often as possible. In most cases when working with a map, we work with tiles array indices, not vectors. SelectionArea also manipulate with indexes (not vectors)

tonihele: I agree using point instead of vectors. Vectors give me the chills. Also LOGIC shouldn't contain jME classes at all (or as little as possible at least). Or was this vector from the Math class or the libGdx... anyway.

ArchDemons:
see #498

PR #386, opened 2021-04-13
Update LICENSE

cxong:
Taken from https://choosealicense.com/licenses/gpl-3.0/

Note: the text near the end are instructions for anyone in general on how to apply GPL3 to their programs, so the template text shouldn't be replaced with OpenKeeper's

Having the license text match the template also helps github auto detect the license and show it in the sidebar


tonihele:
I guess it checks out :) Thanks

cxong:
Unfortunately it hasn't worked; I think the `license.txt` is confusing the license detector

tonihele:
When you view your own branch in the Github, you should have already seen the change if any.

PR #405, opened 2022-01-25
upload build artifacts

Trass3r:
Allows to download the builds from the Actions tab.
https://github.com/Trass3r/OpenKeeper/actions/runs/1745813856

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -25,3 +25,7 @@ jobs:
       run: chmod +x gradlew
     - name: Build with Gradle
       run: ./gradlew build
+    - uses: actions/upload-artifact@v2
+      with:
+        name: distribution
+        path: build/distributions/OpenKeeper.zip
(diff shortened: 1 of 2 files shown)


tonihele:
Cool, is it the latest build or how many builds?

Trass3r:
Every commit, but I think they clean it up after a while.

tonihele:
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy

90 days. So it might be that there is nothing if no activity in 90 days

Trass3r:
Yeah for longer lasting builds there are releases.
Which can also be created via Github Actions.

tonihele:
Yep, we are little bit late with our 0.1 milestone. Still some issues.

But yeah, I'm sure this is all good.

Trass3r:
Already worked:
https://github.com/tonihele/OpenKeeper/actions/runs/1747024306

PR #409, opened 2022-03-30
Animated textures

tonihele:
Implements animating textures, or at least some of them. To be more specific, this is pretty much limited to Fireflies and Fairies. They now have wings!

Also the textures list is actually carried from KMF to J3O. Might come in handy later elsewhere too.

### DIFF
--- a/src/toniarts/openkeeper/view/map/MapViewController.java
+++ b/src/toniarts/openkeeper/view/map/MapViewController.java
@@ -283,15 +283,19 @@ public abstract class MapViewController implements ILoader<KwdFile> {
 
                         // FIXME: This doesn't sit well with the material thinking (meaning we produce the actual material files)
                         // Now we have a random starting texture...
-                        int textureIndex = Math.round((terrain.getTextureFrames() - 1) * (1 - tile.getHealthPercent() / 100f));
+                        int textureIndex = Math.round((terrain.getTextureFrames() - 1) * (tile.getHealthPercent() / 100f));
                         String diffuseTexture = textures.get(textureIndex);
-                        try {
-                            Texture texture = assetManager.loadTexture(new TextureKey(ConversionUtils.getCanonicalAssetKey(diffuseTexture), false));
-                            material.setTexture("DiffuseMap", texture);
+                        String diffuseTextureKey = ConversionUtils.getCanonicalAssetKey(diffuseTexture);
+                        String currentTexture = material.getTextureParam("DiffuseMap").getTextureValue().getKey().getName();
+                        if(!diffuseTextureKey.equals(currentTexture)) {
+                            try {
+                                Texture texture = assetManager.loadTexture(new TextureKey(diffuseTextureKey, false));
+                                material.setTexture("DiffuseMap", texture);
 
-                            AssetUtils.assignMapsToMaterial(assetManager, material);
-                        } catch (Exception e) {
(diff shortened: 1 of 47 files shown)


src/toniarts/openkeeper/view/map/MapViewController.java
```diff
@@ -276,31 +277,26 @@ public void visit(Spatial spatial) {
...
-                if (terrain.getFlags().contains(Terrain.TerrainFlag.DECAY) && terrain.getTextureFrames() > 1) {
+                if (terrain.getFlags().contains(Terrain.TerrainFlag.DECAY)) {
 
-                    Integer texCount = spatial.getUserData(KmfModelLoader.MATERIAL_ALTERNATIVE_TEXTURES_COUNT);
-                    if (texCount != null) {
+                    List<String> textures = spatial.getUserData(KmfModelLoader.MATERIAL_ALTERNATIVE_TEXTURES);
+                    if (textures != null) {
 
                         // FIXME: This doesn't sit well with the material thinking (meaning we produce the actual material files)
                         // Now we have a random starting texture...
```
Trass3r, line 286: comment still applies after the sorting fix?

tonihele: Sort of I guess. The original idea of the whole asset conversion thing was to provide artists access to resources that can be modified. Meaning the material files and all that could be modified in WYSIWYG editor (jME SDK). And close to zero modifications to the assets on runtime, only what is dictated by the level files.

The idea is still fine in my head like this. Just that the WYSIWYG editors for jME format assets are... quite primitive for artists. Modifying models in said editors... very limited. So in reality this idea doesn't really work all that well.

The material conversion of course offers developers easy peek to what is in them, greatly speeding up the development and not needing to craft special tools to take a look. The model conversion for that matter is just waste of time and space. They would be faster to just load from KMFs.

These are the thoughts behind this. I hope it makes some sense.

Trass3r: > The material conversion of course offers developers easy peek to what is in them,

Yeah 👍🏼

> The model conversion for that matter is just waste of time and space. They would be faster to just load from KMFs.

Really? Interesting.

tonihele: The KMFs are max few hundreds of Kbs in size whereas J3Os are around ten times larger. This is from where my claim comes. The process of making KMF readable (small disk read, some CPU) in the engine is not too heavy in comparison to reading it ready from a J3O (big disk read, maybe less CPU). This is largely based on assumptions.

The code however is written so that we could just switch to reading KMFs straight. The ModelViewer, if I remember correctly, already does it like this.

Trass3r: > Just that the WYSIWYG editors for jME format assets are... quite primitive for artists. Modifying models in said editors... very limited.

There's this though: https://github.com/stephengold/Maud

tonihele: That and the SDK (https://github.com/jMonkeyEngine/sdk/releases) which I use. These can be used to some extend. Just that the animation player, the PoseTrack thing is custom and doesn't play in 3rd party tools

Trass3r: Ah yeah makes sense.

Issue #362, opened 2020-04-05
Migrate to the new animation system


tonihele:
Migrate to the new animation system. It supposedly comes with some goodies, and our morph animations are supported out of the box as well as hardware accelerated (?). Should be worthwhile to check it out, at least to get rid of the warnings.

Some write up about the new animation system:
https://hub.jmonkeyengine.org/t/monkanim-v2/39877

Also the place where the magic is currently done is in the KmfModelLoader.

tonihele:
Not as I would have hoped.... I assumed we can use the new morph animation & hardware acceleration straight... But that is a no-go. The morph animation supports kinda whole poses, and you are not supposed to have a lot of them. Up to 14 (well simplification, the rules are a bit more complicated than this...) could be hardware accelerated via the default shader that we are also using.

The problem is that we have a lot of poses in a way. The DK 2 system is very fine grained, it animates specific subsets of vertices between the frames. Thus creating a huge number of complete poses if you translate them to complete poses that is. I tried with the Fairy dancing and it went to 1700 poses (start and end both). That is a no-go. One could with very complex algorithm to try to compress these to smaller amount, combine etc.. But that alone would also be way more than 14 poses and bloat the saved binary.

If you profile OpenKeeper, this pose animation running is what we mostly do, so that is why I'm a bit disappointed. But yeah, we should anyway port our Pose animation to the new animation system anyway.

Trass3r:
What does that `AnimMigrationUtils.migrate(model)` do then?
https://hub.jmonkeyengine.org/t/pose-animation-deprecated/42031

tonihele:
It doesn't, right? https://github.com/jMonkeyEngine/jmonkeyengine/blob/master/jme3-core/src/main/java/com/jme3/anim/util/AnimMigrationUtils.java.

Also does what? :D The only problem I encountered is that the support for the LoopMode is non-existent and it can't be emulated properly due to a bug/missing feature. 

The vertex animation? That I have thought sometimes that if we could do our own Fragment shader. We could feed the animation data as it is in the KMF file (it is packed very nicely) and just have it play on the GPU itself rather than spending a lot of time feeding vertex data to GPU on each frame... But this is just a theory. I have no idea how much memory that would use on GPU and does it really benefit...

Trass3r:
Ah yeah true looks that migration thing just handles skeletal animation.

I don't really understand the kmf animation system. It's too complicated. Is there any write-up about it?
[...]

PR #412, opened 2022-07-11
 Migrate to the new animation system

Trass3r:
Found this old #362 branch and rebased fwiw.

### DIFF
--- a/src/toniarts/openkeeper/animation/PoseTrack.java
+++ b/src/toniarts/openkeeper/animation/PoseTrack.java
@@ -31,25 +31,19 @@
  */
 package toniarts.openkeeper.animation;
 
-import com.jme3.animation.AnimChannel;
-import com.jme3.animation.AnimControl;
-import com.jme3.animation.ClonableTrack;
-import com.jme3.animation.Track;
+import com.jme3.anim.MorphTrack;
 import com.jme3.export.InputCapsule;
 import com.jme3.export.JmeExporter;
 import com.jme3.export.JmeImporter;
 import com.jme3.export.OutputCapsule;
 import com.jme3.export.Savable;
+import com.jme3.math.FastMath;
 import com.jme3.math.Vector3f;
 import com.jme3.scene.Geometry;
 import com.jme3.scene.Mesh;
-import com.jme3.scene.Node;
-import com.jme3.scene.Spatial;
 import com.jme3.scene.VertexBuffer;
 import com.jme3.scene.VertexBuffer.Type;
 import com.jme3.util.BufferUtils;
-import com.jme3.util.TempVars;
-import com.jme3.util.clone.Cloner;
 import java.io.IOException;
 import java.nio.FloatBuffer;
 import javax.annotation.Nullable;
@@ -57,11 +51,9 @@ import javax.annotation.Nullable;
 /**
  * A single track of pose animation associated with a certain mesh.
  */
-public final class PoseTrack implements Track, ClonableTrack {
+public final class PoseTrack extends MorphTrack {
 
-    private int targetMeshIndex;
     private PoseTrack.PoseFrame[] frames;
-    private float[] times;
 
     public static class PoseFrame implements Savable, Cloneable {
 
@@ -85,7 +77,7 @@ public final class PoseTrack implements Track, ClonableTrack {
          * @return a clone of the current object
          */
         @Override
-        public PoseTrack.PoseFrame clone() {
+        public PoseTrack.PoseFrame clone() throws CloneNotSupportedException {
             try {
                 PoseTrack.PoseFrame result = (PoseTrack.PoseFrame) super.clone();
                 result.weights = this.weights.clone();
@@ -121,9 +113,9 @@ public final class PoseTrack implements Track, ClonableTrack {
         }
     }
 
-    public PoseTrack(int targetMeshIndex, float[] times, PoseTrack.PoseFrame[] frames) {
-        this.targetMeshIndex = targetMeshIndex;
-        this.times = times;
+    public PoseTrack(Geometry target, float[] times, PoseTrack.PoseFrame[] frames)
+    {
+        super(target, times, new float[times.length], 1);
         this.frames = frames;
     }
 
@@ -131,22 +123,6 @@ public final class PoseTrack implements Track, ClonableTrack {
      * Serialization-only. Do not use.
      */
     public PoseTrack() {
-    }
-
-    @Override
-    public float[] getKeyFrameTimes() {
-        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
-    }
-
-    @Override
-    public Track cloneForSpatial(Spatial spatial) {
-        PoseTrack track = new PoseTrack(targetMeshIndex, times, frames);
-        return track;
-    }
-
-    @Override
-    public void cleanUp() {
-
     }
 
     private void applyFrame(Mesh target, int frameIndex) {
@@ -184,10 +160,7 @@ public final class PoseTrack implements Track, ClonableTrack {
             Vector3f endOffset = endPose.getOffsets()[i];
 
             // Interpolate
-            interpOffset.set(endOffset);
-            interpOffset.subtractLocal(startOffset);
-            interpOffset.multLocal(weight);
-            interpOffset.addLocal(startOffset);
+            FastMath.interpolateLinear(weight, startOffset, endOffset, interpOffset);
 
             // Write modified vertex
             BufferUtils.setInBuffer(interpOffset, vertexBuffer, vertIndex);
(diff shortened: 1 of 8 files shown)


src/toniarts/openkeeper/tools/modelviewer/ModelViewer.java
```diff
@@ -582,11 +580,12 @@ private void setupModel(final Node spat, boolean isMap) {
...
-                if (animControl != null) {
-                    AnimChannel channel = animControl.createChannel();
-                    channel.setAnim("anim");
-                    AnimationLoader.setLoopModeOnChannel(spatial, channel);
+                var animComposer = spatial.getControl(AnimComposer.class);
+                if (animComposer != null) {
+                    animComposer.setCurrentAction("anim");
+                    // var channel = animControl.getAction(NODE_NAME).createChannel();
+                    // channel.setAnim("anim");
+                    // AnimationLoader.setLoopModeOnChannel(spatial, channel);
```
Trass3r, line 588: ?

Trass3r: > The only problem with it that it doesn't natively support the loop modes. And it can't be (at the time at least) extended to support them as there is a bug/feature in there that prevents it.

I see, maybe it's fixed by now?

tonihele: https://github.com/jMonkeyEngine/jmonkeyengine/pull/1038 this at least was never merged to the engine... I was a bit lazy on this, should have pursued it further... But maybe there is some way. I just stopped because I wanted to wait for this. Never looked elsewhere.

Trass3r: I see, what about https://github.com/jMonkeyEngine/jmonkeyengine/pull/1038#issuecomment-792282131?

tonihele: I think the problem was something that while we could play it backwards on the second loop, we couldn't catch that anymore ending (to have it again looping but this time forward again). It was some time ago. ModelViewer is anyway the easiest to catch these. Some Imp work animations, if I remember correctly, had different loop modes to test on.

src/toniarts/openkeeper/animation/PoseTrack.java
```diff
@@ -57,9 +52,9 @@
 /**
  * A single track of pose animation associated with a certain mesh.
  */
-public final class PoseTrack implements Track, ClonableTrack {
+public final class PoseTrack extends MorphTrack {
```
Trass3r, line 55: I guess this could be a first step towards solving https://github.com/tonihele/OpenKeeper/pull/409#discussion_r918624619.
AFAICT you'd need to define `MorphTarget`s for the target Geometry instead of custom `Pose`s?

tonihele: The jME morph targets work a bit differently. If I remember correctly, they are hardware accelerated. But there was some reason that they don't really work with our way of thinking... Was it that we would then have millions of morph targets per model? And it just didn't work out... Something like this? And the hardware acceleration worked only up until 7 morph targets...

Trass3r: Yeah I think it's similar to what I do in blender, 1 morph target per frame and weights like (0,0,0,1,0,0...), so quite a few.
But maybe MorphTrack could otherwise be abused into what we need.

https://github.com/jMonkeyEngine/jmonkeyengine/blob/af375391ea385a061444ca03262e1a4720c3274d/jme3-core/src/main/java/com/jme3/anim/tween/action/ClipAction.java#L49-L53

src/toniarts/openkeeper/animation/PoseTrack.java
```diff
@@ -57,9 +52,9 @@
 /**
  * A single track of pose animation associated with a certain mesh.
  */
-public final class PoseTrack implements Track, ClonableTrack {
+public final class PoseTrack extends MorphTrack {
 
-    private int targetMeshIndex;
+    private Geometry target;
     private PoseTrack.PoseFrame[] frames;
     private float[] times;
```
Trass3r, line 59: I guess this is now redundant with
https://github.com/jMonkeyEngine/jmonkeyengine/blob/84d7d3bb2ddc434a65ef06ac789c245d21e1e359/jme3-core/src/main/java/com/jme3/anim/MorphTrack.java#L56

src/toniarts/openkeeper/animation/PoseTrack.java
Trass3r: Thanks!

The problem is that the whole PoseTrack thing is even harder to understand than KMF itself. All those previous and next frame refs in FrameInfo and Pose, all the maps and copy paste calls in:
https://github.com/tonihele/OpenKeeper/blob/604a33a377c6c069c7cf60dfea3bdaa7711b9899/src/toniarts/openkeeper/tools/convert/KmfModelLoader.java#L414
There also seems to be quite some redundancy:
https://gist.github.com/Trass3r/cdd614ec24c48f94969389492117fa0d#file-piranha_swim-xml-L163

Trass3r: Looks like we were already passing through this code, it works cause in `getDataAtTime` we never set the output weights parameter and then `setMorphState` is a no-op:
```java
    private void interpolateMorphTrack(double t, MorphTrack track) {
        Geometry target = track.getTarget();
        float[] weights = target.getMorphState();
        track.getDataAtTime(t, weights);
        target.setMorphState(weights);
```

Btw the shader is indeed limited to a handful of morph targets to blend but you don't really need that many, and the system also falls back to CPU morph target merging as needed:
https://github.com/jMonkeyEngine/jmonkeyengine/blob/af375391ea385a061444ca03262e1a4720c3274d/jme3-core/src/main/java/com/jme3/anim/MorphControl.java#L157-L158

src/toniarts/openkeeper/tools/convert/KmfModelLoader.java
```diff
@@ -490,28 +502,20 @@ private VertexBuffer[] createIndices(final Map<Integer, List<Triangle>> triangle
...
                 for (Triangle triangle : triangles.getValue()) {
                     indexes[x * 3] = triangle.getTriangle()[2];
                     indexes[x * 3 + 1] = triangle.getTriangle()[1];
                     indexes[x * 3 + 2] = triangle.getTriangle()[0];
-                    x++;
+                ++x;
                 }
-                VertexBuffer buf = new VertexBuffer(Type.Index);
-                buf.setupData(VertexBuffer.Usage.Static, 3, VertexBuffer.Format.UnsignedInt, BufferUtils.createIntBuffer(indexes));
+            buf.setupData(VertexBuffer.Usage.Static, 3, VertexBuffer.Format.UnsignedShort, BufferUtils.createShortBuffer(indexes));
```
Trass3r, line 503: The original data is even in bytes, not sure why it's saved as shorts.
https://github.com/tonihele/OpenKeeper/blob/995a56ea9e3b2eabe02614ff676fec4a7e33fe3f/src/toniarts/openkeeper/tools/convert/kmf/KmfFile.java#L340-L342

tonihele:
What is the status of this? I didn't try the code yet, do the different animation modes still work?

Trass3r:
So far I couldn't spot any regressions.
But please feel free to test it yourself.

tonihele:
On 3.6 jME there will be native tweens for the deprecated loop modes. So lets look at this then.

Trass3r:
I see.
https://github.com/jMonkeyEngine/jmonkeyengine/releases/tag/v3.6.0-alpha1

Trass3r:
Would we need to use AnimLayer then?
https://github.com/jMonkeyEngine/jmonkeyengine/pull/1845

tonihele:
Hmm, not sure, it was this PR https://github.com/jMonkeyEngine/jmonkeyengine/pull/1849/files.

Trass3r:
> Some enhancement to new animation system by @Ali-RS in https://github.com/jMonkeyEngine/jmonkeyengine/pull/1845

> Added a Loop tween to Tweens factory class. Supports looping by count… by @Ali-RS in https://github.com/jMonkeyEngine/jmonkeyengine/pull/1846

> Added Tweens.cycle() and Tweens.invert() methods by @Ali-RS in https://github.com/jMonkeyEngine/jmonkeyengine/pull/1849

Trass3r:
It has landed: https://github.com/jMonkeyEngine/jmonkeyengine/releases/tag/v3.6.0-stable

src/toniarts/openkeeper/tools/convert/KmfModelLoader.java
tonihele: Oh yeah, sorry. I think I answered in the other PR just now. But yeah, lets put it here too. We read everything as one size bigger than the original to count for the unsigned stuff. That is how we get correct value to the file format objects. How it is then used in our conversions... that is another thing.

tonihele:
Yeah, I finally got around to test this. It seems to work.

I tried to use the Tween.cycle in the AnimationLoader rather than tamper with the Global Animation Speed. But that resulted in doors spawning as opened. Well, need to check this at some point.

So yes, fine by me. Ready to merge?

PR #421, opened 2022-10-02
Use the new Java Date/Time API

tonihele:
Use the _new_ Java Date/Time API... from Java 8. Sorry for being such a late bloomer :)

Tasks now use Instant rather than Date. This is actually a kinda bug fix even. The tasks could even go broke when automatically changing DST. Or in the future loading a saved game on another timezone, or just changing your timezone... Well, you get the drift, using date here is just stupid, always was.

What comes to KWD timestamps... This has been broken all along.

- We assumed that the timestamp is in UTC. I tried this with the map editor. The timestamp is always the local time. And there is no timezone information (that I'm aware of). So we have no way of actually getting accurate timestamp.
- The months actually are from 1-12. With the old API 0-11 is assumed. The month was always off by one.
- There is no timestamp available on all files. We now reflect this and parse it as null rather than giving it some value.

### DIFF
--- a/src/toniarts/openkeeper/game/task/AbstractTask.java
+++ b/src/toniarts/openkeeper/game/task/AbstractTask.java
@@ -19,7 +19,7 @@ package toniarts.openkeeper.game.task;
 import com.jme3.math.Vector2f;
 import com.simsilica.es.EntityId;
 import java.awt.Point;
-import java.util.Date;
+import java.time.Instant;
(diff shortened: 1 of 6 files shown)


tonihele:
DeepSource is right... It complains about the exceptions (using just the RuntimeException everywhere). We don't really have any of our own. Which... yes, we should have.

PR #424, opened 2022-12-11
Improve lookup

tonihele:
Use matrix instead of map for the storing the entity positions and pre-init all the matrix locations for constant access times. Also decreases the code indentation in places.


tonihele:
Yeah, DeepSource is again right. We should tackle the exceptions at some point....

tonihele:
Hmm, GH bug. This is merged but the status is all wrong. Github has a concurrency bug I'm sure...

PR #427, opened 2023-03-28
Make sure loggers are initialized first

tonihele:
Fixes a bug when the folder is configured, but not valid. It actually crashes to the LOGGER being null. As a static variable listed earlier than the logger causes a handled crash. But the handling fails....

So make sure we declare logger as the first variable to avoid these by accident.


tonihele:
Deepsource didn't find any new issues. Valid still... But maybe another time.

PR #428, opened 2023-04-13
upgrade to JME 3.6

Trass3r:
Everything seems to work as before, but for some reason the settings dialog doesn't come up when you launch the model viewer.


tonihele:
> but for some reason the settings dialog doesn't come up when you launch the model viewer.

It needs this:
https://mvnrepository.com/artifact/org.jmonkeyengine/jme3-awt-dialogs

It was made separate package since AWT + LWJGL 3 (GLFW really) is a poisonous combination on MacOS and Linux. Nothing has really changed, it has always been like this. So we can just have the package anyway. We even have our own AWT dialogs...

Trass3r:
Thanks! Added it.

tonihele:
Thanks for the contribution!

PR #431, opened 2023-04-15
refactor index buffers creation

Trass3r:


src/toniarts/openkeeper/tools/convert/kmf/KmfFile.java
```diff
@@ -333,16 +332,22 @@ private List<MeshSprite> parseMeshSprites(IResourceChunkReader rawKmf, int sprsC
...
+                for (int k = 0; k < numTris; ++k) {
                     triangles.add(new Triangle(rawKmf.readUnsignedByte(),
                             rawKmf.readUnsignedByte(),
                             rawKmf.readUnsignedByte()));
                 }
-                trianglesPerLod.put(j, triangles);
+                trianglesPerLod.add(triangles);
             }
+
+            // remove trailing empty LODs
```
Trass3r, line 347: I guess that's ok looking at
https://github.com/jMonkeyEngine/jmonkeyengine/blob/76ebd1443ccc18249e9f082d528cd39baf59da84/jme3-core/src/main/java/com/jme3/scene/control/LodControl.java#L179-L183

Trass3r: Gem holder looks normal in the model viewer so not sure if it breaks something there.

src/toniarts/openkeeper/tools/convert/kmf/Triangle.java
```diff
@@ -23,16 +23,17 @@
  */
 public class Triangle {
 
-    private final short[] triangle;
+    private final byte[] triangle;
```
tonihele, line 26: Hmm, this doesn't break anything? We read everything one size bigger than the actual file format since Java doesn't have unsigned (well, ok, nowadays it has but those are, well, never used but looks cumbersome...).

AssetsConverter.ConvertProcess.Models I believe requires an update. This will trigger the automatic conversion.

Also if at all possible. I would prefer to have the LOD cleaning done in KmfModelLoader instead. As the files in these formats, like under package `kmf`, are supposed to be as close as possible to the original format. Raw format. Read with all, or most, of the garbage they might include. Then `KmfModelLoader `is where we interpret this data, and that is where we have to do all the hacking.

This is to try also to serve the public by keeping the file formats clear.

Good job!! Thank you so much!

src/toniarts/openkeeper/tools/convert/kmf/KmfFile.java
Trass3r: Though thinking again about how that code expects a sorted set of numbers I'm not sure why it works at all, resp. doesn't fail horribly.

src/toniarts/openkeeper/tools/convert/kmf/Triangle.java
Trass3r: If you used the bytes directly in Java you'd get negative numbers for the larget values. But ee pass them on to the index buffers and GL supports unsigned Interpretation.

Do you always increment AssetsConverter versions to avoid confusion about code vs cache or only when it's a breaking change to reduce the number of reprocessings?

Good point about the loader, will do.

tonihele: It is a good practice to increase it always when a loader changes. From here I can't determine is there an actual change in the models or not. Then, like you said, code matches the _cache_. Also good to monitor that nothing breaks and if the file sizes suddenly increase drastically... is there going to be a performance hit etc.

This is development stage. I wouldn't worry about the asset conversion process. I have Skylake era Intel processor and all that, it is just 2 minutes to do the complete conversion. And increasing just the model version number will just trigger model conversion. One time.

tonihele:
Looks good to me. Is this only to fix the tangent generation (and bad data)? Was there anything else to this?

Trass3r:
It started simple by noticing we create int32 index buffers even though the source data is only int8.
But it opened the whole can of worms with the LOD system etc.

So I guess I'll try to keep this small as a cleanup that shouldn't change any observable behavior.

PR #442, opened 2023-10-21
Bugfix/rendering fixes

tonihele:
Enables our animated texture material to work with single pass light mode. Also fixes some shadow modes.


tonihele:
Yeah, I'll create an issue about the toniarts.openkeeper.video.tgq.TgqFile.readAudioHeader(int frameSize). Looks kind of a daunting method.

PR #445, opened 2023-11-30
clean up some of the deprecated code usages

Trass3r:


tonihele (approved):
Looks good to me. But exposes the problem that we reference the view from the core. But that was there all along.

I'll probably chop this to subprojects at some point to make the architecture more clear....

Ready for merge?

tonihele:
Thank you for your contribution!

PR #447, opened 2023-12-16
Feature/java21

tonihele:
Java 21, yay!


tonihele:
Looks like DeepSource goes up to Java 19 only according to its documentation....

Issue #218, opened 2016-08-12
Android compatibility


Trass3r:
JME itself has good Android support: http://docs.jmonkeyengine.org/jme3/android.html
Are there any long-term plans to work towards Android compatibility?

tonihele:
No, not at the moment. Primary focus is on Win/Linux/Mac, well, PC. Enhance the experience and add some spark to it, the modern goodies. And that is kinda working away from Android. But not ruling it out of course.

I've no objection if somebody picks up this feature. As long as it doesn't hinder the primary platform development. I'm still on my Samsung S2, so no help from me :)

Trass3r:
I see. As far as I know one would have to get away from awt/swing. Not sure if there's more regarding the core game, i.e. excluding tools etc.

tonihele:
We're not using awt/swing, as far as I know. Changes are OpenKeeper would start now with Android, but probably too hard to use, if it doesn't crash before.

Trass3r:
When searching the repo it seems to use JFrame in the main and a bunch of awt Events, Points, Graphics etc.

tonihele:
Ah oh yes those. If Points need to replaced, that would be really simple. Swing is for the setup, but that is one thing I don't know how Android should do it really (we ask for the game directory).

ufdada:
Well technically we could do it ingame, but on the other hand i think it's not really pretty to choose an installation-folder in android. We should think about a smarter way on other plattforms than windows anyway.

Trass3r:
https://github.com/search?q=repo%3Atonihele%2FOpenKeeper%20awt&type=code
https://github.com/tonihele/OpenKeeper/issues/304#issuecomment-352272004

PR #456, opened 2024-06-16
Feature/graphics options

tonihele:
Fixes two things.

I installed a fresh system and saw that the gamma correction is on by default, came with some jME upgrade. So unless you have old settings, your videos and graphics are a bit burnt. Too bright. Turn gamma off by default.

The current way of getting display modes is really bugged. Especially on Linux, it is almost impossible to get correct readings. Also with MacOS and LWJGL 3 this poses an actual bug that causes freeze. We now query the display modes primarily from LWJGL (2 & 3 both supported), this should give much better accuracy than the current AWT (used as a fallback only with this PR).

### DIFF
--- a/src/toniarts/openkeeper/game/data/Settings.java
+++ b/src/toniarts/openkeeper/game/data/Settings.java
@@ -267,10 +267,6 @@ public class Settings {
         // Init the settings
         this.settings = settings;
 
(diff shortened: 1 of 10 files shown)


Trass3r:
Also good for #218 👍🏽

PR #460, opened 2024-07-08
Feature/misc improvements

tonihele:
- Imps drop the excess gold when digging at their feet if there is no place to store it currently. Prevents imps from coming millionaires
- Creature fleeing is a bit nicer now, always annoyed me the they could not decide what to do, now at least they are a bit more decisive and stay fleeing a minimum time

### DIFF
--- a/src/toniarts/openkeeper/game/controller/creature/CreatureController.java
+++ b/src/toniarts/openkeeper/game/controller/creature/CreatureController.java
@@ -1136,29 +1136,38 @@ public class CreatureController extends EntityController implements ICreatureCon
     @Override
     public boolean isStateTimeExceeded() {
         double timeSpent = gameTimer.getGameTime() - entityData.getComponent(entityId, CreatureAi.class).stateStartTime;
+        double stateTargetTime;
 
         switch (stateMachine.getCurrentState()) {
-            case STUNNED: {
+            case STUNNED -> {
                 // Hmm, this might actually be the level variable, the stun seems to be the time fallen when dropped
-                return timeSpent >= entityData.getComponent(entityId, CreatureComponent.class).stunDuration;
+                stateTargetTime = entityData.getComponent(entityId, CreatureComponent.class).stunDuration;
             }
-            case FALLEN: {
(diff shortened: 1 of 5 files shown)


tonihele:
I don't actually have now the original game at hand (Linux...) so I can't check whether the gold clumped up to bigger piles (the max pile size) on the same tiles or just like this. A lot of small clumps. This has an actual effect on the game as the imp tasks are not grouped. The smaller the piles, the less effective your imps are collecting it...

src/toniarts/openkeeper/game/controller/creature/CreatureController.java
```diff
@@ -858,7 +858,21 @@ public void resetReEvaluationTimer() {
     @Override
     public void addGold(int amount) {
         Gold gold = entityData.getComponent(entityId, Gold.class);
-        entityData.setComponent(entityId, new Gold(gold.gold + amount, gold.maxGold));
+
+        // Drop excess gold, we can only carry so much
+        boolean hasMaxGold = gold.maxGold > 0;
+        int maxGoldCanAdd = hasMaxGold ? gold.maxGold - gold.gold : Integer.MAX_VALUE;
```
Trass3r, line 864: Why isn't MAX_VALUE used to designate no limit in the first place?

tonihele: That is a question I ask myself every night I go to sleep... :D Yeah, I'll change it. The logic is still a bit iffy here anyway, I didn't even check that do normal creatures have a max value they can carry? Essentially dropping all bribes and pay they are given if they didn't have a chance to blow it all in the rigged casino...

tonihele: Ok, I checked. Creatures always have some max gold set, everybody. 0 is not used to indicate max gold. That was my half assed assumption. So removed all assumptions from here.

Also normal creatures do have max gold. I guess it makes sense in the real world. But I don't know how the normal creatures deal with their possessions...

PR #469, opened 2024-09-27
add automatic 'latest' release

Trass3r:
Example: https://github.com/Trass3r/OpenKeeper/releases


Trass3r:
If you merge, please squash.

tonihele:
Hmm, is there a cleanup for the releases? Every PR merge will now create an actual release. And those will just pollute any actual releases? If it only holds the latest, then it is acceptable but we don't need an archive (we have a version control system).

I'm just thinking who is the target audience for this? We don't do much with consumers as this is not a consumer ready product yet. And developers are able to do just compile this themselves.

Trass3r:
Yeah it just maintains a 'latest' release.
Just to make it a bit more accessible to try out for new users.

PR #481, opened 2025-02-04
Gradle to 8.12.1

tonihele:


tonihele:
I had some trouble building OpenKeeper after getting the latest. Some local problem as the CI seems to be ok. I tried this Gradle version. Didn't make a difference, but since I did... I might as well push it.

PR #491, opened 2025-08-03
Misc fixes

Trass3r:


src/toniarts/openkeeper/game/console/GameConsole.java
```diff
@@ -169,7 +169,10 @@ public void execute(final String[] args) {
                 case ADD_GOLD:
                     try {
                         int amount = Integer.parseInt(args[1]);
-                        stateManager.getState(WorldState.class).addGold(keeper.getId(), amount);
+                        stateManager.getState(GameClientState.class)
```
tonihele, line 172: This does nothing but corrupts the client state. The moneys are kept on the server side.

tonihele: throw NotImplementedException()? :D I mean, it would probably require its own interface for these debug methods. `GameClientService` does have `triggerCheat` already. Maybe something similar then for debug functions, at least they should be grouped together for easy identification and validation.

tonihele (approved):

PR #498, opened 2025-08-28
build & sell system

ArchDemons:


src/toniarts/openkeeper/game/logic/BuildTileSystem.java
```diff
@@ -0,0 +1,120 @@
...
+        this.gameWorldController = gameWorldController;
+        entities = entityData.getEntities(TileBuildOrSell.class);
+    }
+
+    @Override
+    public void processTick(float tpf) {
+        // Add new & remove old
+        entities.applyChanges();
+
+        synchronized (this.queue) {
```
tonihele, line 53: Locks? I'm trying to get rid of all locks... We still have the gold lock but that too should go.

src/toniarts/openkeeper/game/controller/GameWorldController.java
```diff
@@ -343,14 +334,12 @@ public void build(Vector2f start, Vector2f end, short playerId, short roomId) {
             return;
         }
 
-        // If this is a bridge, we only got the starting point(s) as valid so we need to determine valid bridge pieces by our ourselves
         Room room = kwdFile.getRoomById(roomId);
         if ((room.getFlags().contains(Room.RoomFlag.PLACEABLE_ON_WATER))
                 || room.getFlags().contains(Room.RoomFlag.PLACEABLE_ON_LAVA)) {
-            instancePlots = new ArrayList<>(mapController.getTerrainBatches(instancePlots, x1, x2, y1, y2));
+//            instancePlots = new ArrayList<>(mapController.getTerrainBatches(instancePlots, x1, x2, y1, y2));
```
tonihele: We should just remove the code, not comment it, a bit confusing. Version control is there to manage these.

build.gradle
```diff
@@ -64,6 +65,8 @@ dependencies {
     implementation "com.simsilica:zay-es-net:1.6.0"
     implementation "com.simsilica:sio2:1.8.0"
     implementation "com.simsilica:sim-ethereal:1.8.0"
+    implementation "com.google.guava:guava:33.4.8-jre"
+    implementation "org.reflections:reflections:0.10.2"
```
tonihele, line 69: I've been removing these (reflections) at work :D For security reasons. Are there any alternatives?

tonihele:
Generally the idea is fine. Lot of trivial DeepSource issues. Still WIP?

build.gradle
ArchDemons: We need scan package for classes with annotation... I don`t know any other library that can do this

src/toniarts/openkeeper/game/logic/BuildTileSystem.java
ArchDemons: maybe we don`t need use lock there. Doesn`t check thead safe without lock of this code

build.gradle
tonihele: Well, me neither. I just designed around it at work. It is a bit of a hazard in my opinion to take this in. Is there any alternative to having to scan package classes with annotation? I know what you are trying to solve here, having to manually add the class to be de/serialized. How else we could do it better?

src/toniarts/openkeeper/game/controller/GameWorldController.java
ArchDemons: we need this code :), but not compatible with new code. This is the reason why the PR is in WIP

ArchDemons:
I want to try something lik Command-Query on Event Bus. Each client send query (e.g. i want build room 5x5) to server. Server checks possibility, and send command to client (e.g. build room 5x5) without any checks on client

build.gradle
ArchDemons: compile-time annotation preprocessor (like lombok), but it`s rocket science for me :). I work with spring framework. Packages scan - the core of DI container and dynamic connection of the components

src/toniarts/openkeeper/game/logic/BuildTileSystem.java
tonihele: Yeah, the ECS system we use is thread safe. And all commands coming from client I planned to enqueue to the main game loop thread. That way we don't need to mess about with locks (the gold lock or any other). I'm unsure though how would this affect the game as there would be a small delay in processing the client commands.

PR #508, opened 2026-08-01
490: Replace text parameter indices with enum values

### LINKED ISSUE #490: Text replacement should use enum values instead of magic integers
Text replacement should use enum values instead of magic integers. `TextUtils` parse the indices (the magic integers) from the strings and pass these to the text replacer. It would be incredibly helpful and easy to use enum values instead. The magic numbers seem to mean the same thing in all of the strings. And the way we currently do this is scattered to different handlers that know the context. Meaning of the magic integers are duplicated in all of those.

Some values we already have mapped in the code, and some would be helpful to extract from the translation files. These are usually simple enough to be deducted from the context.

belgianguy:
Fixes #490 (conservative approach, just the enum)

### DIFF
--- a/src/main/java/toniarts/openkeeper/utils/TextParameter.java
+++ b/src/main/java/toniarts/openkeeper/utils/TextParameter.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright (C) 2014-2026 OpenKeeper
+ *
+ * OpenKeeper is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ */
+package toniarts.openkeeper.utils;
+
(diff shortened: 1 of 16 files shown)


tonihele (approved):

Trass3r:
```
WARNING toniarts.openkeeper.utils.TextUtils getUnsupportedParameterMessage - Text parameter 41 is not supported by RoomTextParser
WARNING toniarts.openkeeper.utils.TextUtils getUnsupportedParameterMessage - Text parameter 40 is not supported by RoomTextParser
```

belgianguy:
@Trass3r : Thanks, those are the current and maximum Dungeon Heart mana parameters. The enum mapping is correct, but RoomTextParser currently only receives room information and therefore cannot resolve them. The new warning exposed an existing unimplemented replacement. I’ll add player mana context to the room parser in a follow-up issue.

PR #516, opened 2026-08-11
Preserve depth for models held in keeper hand

### LINKED ISSUE #515: Held models lose depth in the Hand of Evil, causing Bile Demon attachments to clip
### Problem

Multipart 3D models can render incorrectly while held in the Hand of Evil.

The clearest example is the Bile Demon: its ear balls and spikes are partially
covered by the rear body mesh while playing `Bile_Demon-Swing`. The same raw
animation renders correctly in ModelViewer and in Dungeon Keeper II.

### Cause
`KeeperHandState` currently attaches the held 3D model beneath jMonkeyEngine's

`guiNode`, so it inherits `RenderQueue.Bucket.Gui`.

jMonkeyEngine renders the GUI bucket with its depth range collapsed to `(0, 0)`.
All fragments of the held model therefore receive the same depth. Because the
default comparison is `LessOrEqual`, later KMF submeshes overwrite earlier ones.

For `Bile_Demon-Swing`, the submesh order is:

1. Front body
2. Ear attachments
3. Rear body

The rear body is submitted last and consequently paints over the ear
attachments wherever they overlap on screen.

### Expected behavior

Held 3D models should retain their internal per-pixel depth, as they do in
ModelViewer and Dungeon Keeper II.

### Proposed fix

Render the held 3D model in a dedicated depth-preserving viewport between the
dungeon and GUI passes:

- Use a pixel-aligned orthographic camera.
- Clear depth, but preserve the existing color buffer.
- Keep cursor and queue icons in the normal GUI layer.
- Leave pickup, drop, visibility, and input behavior unchanged.
- Remove the viewport during `KeeperHandState` cleanup.

### Verification

Tested with the Bile Demon in `EnvMapShowcase`:

- Both ear attachments render without clipping while held.
- Dropping the creature works normally.
- Picking it up again works normally.
- Cursor and queue behavior remain unchanged.

Before:

Image

After:

Image

belgianguy:
Fixes #515

### DIFF
--- a/src/main/java/toniarts/openkeeper/view/KeeperHandState.java
+++ b/src/main/java/toniarts/openkeeper/view/KeeperHandState.java
@@ -27,6 +27,8 @@ import com.jme3.math.FastMath;
 import com.jme3.math.Matrix3f;
 import com.jme3.math.Quaternion;
 import com.jme3.math.Vector3f;
+import com.jme3.renderer.Camera;
+import com.jme3.renderer.ViewPort;
 import com.jme3.scene.Node;
 import com.jme3.scene.Spatial;
 import com.jme3.scene.Spatial.CullHint;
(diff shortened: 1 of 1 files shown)


tonihele:
Thanks!

PR #518, opened 2026-08-12
Add better room building rules

### LINKED ISSUE #517: Match DKII room and bridge placement rules
OpenKeeper currently allows partially valid room selections and can build rooms over invalid tiles such as existing rooms, doors, traps, or objects. Room placement should match Dungeon Keeper II and reject the entire selection when any selected tile is invalid.

## Expected behaviour:

- Invalid selections are shown in red.
- No part of an invalid selection is built.
- Normal rooms can only be built on claimed, non-solid land owned by the player.
- Existing rooms must be sold before another room can replace them.
- Doors, traps, and objects block room construction.
- Creatures do not block room construction.
- Wooden and stone bridges can only be built on water or lava.
- A bridge selection must start adjacent to an owned walkable tile or existing owned bridge.
- Once anchored, bridge length is limited only by available gold and valid connected liquid tiles.
- Wooden bridges over lava disappear after `WOOD_BRIDGE_LIFE_ON_LAVA_SECONDS`—30 seconds in the DKII data.
- Stone bridges over lava remain permanently.
- Client preview and server validation must use the same rules.

belgianguy:
Fixes #517

### DIFF
--- a/src/main/java/toniarts/openkeeper/game/controller/RoomPlacementValidator.java
+++ b/src/main/java/toniarts/openkeeper/game/controller/RoomPlacementValidator.java
@@ -68,7 +68,7 @@ public final class RoomPlacementValidator {
             Set<Point> blockedTiles, Vector2f start, Vector2f end, short playerId,
             short roomId, int availableGold) {
         Room room = kwdFile.getRoomById(roomId);
-        if (room == null || !room.getFlags().contains(Room.RoomFlag.BUILDABLE)) {
+        if (!isBuildableRoom(room)) {
             return invalid(Failure.INVALID_ROOM);
         }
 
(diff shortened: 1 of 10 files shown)


belgianguy:
@tonihele it seems DeepSource flagged the GOLD_LOCK here, but IIRC that cannot be fixed yet (at least not in this PR)
The other 4 "Integer but got 1" seem odd to me, any idea?

Agent suggestion for GOLD_LOCK: 
For a proper fix, encapsulate the loose-gold storage operation behind GameWorldController or another gold coordinator so the shared lock can become private.

tonihele:
Gold lock can be removed once everything is moved to the game loop, i.e. sequentially processed. What is left is user input, both single and multiplayer. Should go through a queue that is processed by the game loop.

src/main/java/toniarts/openkeeper/view/PlayerInteractionState.java
```diff
@@ -756,6 +688,27 @@ private boolean canCastSpell(KeeperSpell keeperSpell, IEntityViewControl object,
         return KeeperSpellCastValidator.isValidCast(keeperSpell, kwdFile, mapInformation, mapInformation.getMapData().getTile(tile), gameClientState.getPlayer(), entityData, object != null ? object.getEntityId() : null);
     }
 
+    private Set<Point> getConstructionBlockingTiles() {
+        constructionBlockingObjects.applyChanges();
```
tonihele: If I understand the library correctly, this is what we should avoid. Meaning updating EntitySets _sometimes_. We will run out of memory. You can double check this but I understood that all the changes will just back up somewhere until `applyChanges()` is called.

src/main/java/toniarts/openkeeper/game/controller/RoomPlacementValidator.java
```diff
@@ -0,0 +1,220 @@
...
+
+        public boolean isValid() {
+            return failure == Failure.NONE;
+        }
+    }
+
+    private RoomPlacementValidator() {
+    }
+
+    public static Result validate(KwdFile kwdFile, IMapInformation<?> mapInformation,
```
tonihele: This at least should never be used by client. Causes round trips to server. I should really separate the client and server to make a clear boundary. Of course the ECS library still allows us to mistakes easily. At the very least this should be Javadocced to prevent usage.

Ideally view package should use its own, server its own, common location can have the static common parts. But yeah, probably needs some architecturing and refactoring to make it smooth

belgianguy: Fixed in 301d9000. I removed the EntityData overload from the shared validator. The client now gathers blockers from its subscribed EntitySets, while GameWorldController independently gathers authoritative blockers from server-local EntityData. Both pass only a Set<Point> into the common static validation logic, so no client-side direct EntityData query or extra server round trip is possible through this API.

src/main/java/toniarts/openkeeper/view/PlayerInteractionState.java
belgianguy: Fixed in 301d9000. The three long-lived EntitySets are now drained on every active update before selection processing; getConstructionBlockingTiles() only snapshots their current contents. This prevents queued ECS changes from accumulating while no construction preview is active.

belgianguy:
Review updates are now in 301d9000, after merging current master (including the creature-generator spawn-timer fix from #510). The four DeepSource Integer/remove findings are fixed with explicit boxing. GOLD_LOCK remains intentionally in place; the moved occurrence has a narrow suppression and explanatory comment because removing the lock safely depends on the separate change to queue all player input through the game loop, as discussed.

tonihele:
Thanks for your contribution!

PR #525, opened 2026-08-19
Campaign progress

### LINKED ISSUE #452: "new campaign" button does not work
Start the game and press "new campaign", it will not do anything except make the 'select' noise.

new-campaign

ufdada:
- Level can now be traversed like in the original game
- Viewable cutscenes get unlocked after a level they would be played after (like in the original)
- Level progress can be reset (also resets the viewable cutscenes, like in the original)
- Only the current level arrow is now showing, also with playing and staying red
- Secret level don't show up if they haven't been discovered (currently not implemented)

Not in scope:
- MPD Level progress (already done?)
- Debriefing (currently buggy as hell)
- Playing videos after completing a level
- different texture for completed levels

fixes #452 (but not all the things i mentioned there)

### DIFF
--- a/src/main/java/toniarts/openkeeper/game/state/GameServerState.java
+++ b/src/main/java/toniarts/openkeeper/game/state/GameServerState.java
@@ -43,7 +43,7 @@ import toniarts.openkeeper.game.state.session.GameSessionServerService;
 import toniarts.openkeeper.game.state.session.GameSessionServiceListener;
 import toniarts.openkeeper.tools.convert.map.Door;
 import toniarts.openkeeper.tools.convert.map.KeeperSpell;
-import toniarts.openkeeper.tools.convert.map.KwdFile;
+import toniarts.openkeeper.tools.convert.map.IKwdFile;
 import toniarts.openkeeper.tools.convert.map.Room;
 import toniarts.openkeeper.tools.convert.map.Trap;
 import toniarts.openkeeper.utils.Utils;
@@ -66,7 +66,7 @@ public final class GameServerState extends AbstractAppState {
     private volatile boolean gameLoaded = false;
 
     private final KwdFile kwdFile;
-    private final toniarts.openkeeper.game.data.CampaignLevel levelObject;
+    private final toniarts.openkeeper.game.data.Level levelObject;
 
     private final boolean campaign;
     private final boolean multiplayer;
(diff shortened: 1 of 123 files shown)


src/main/java/toniarts/openkeeper/game/state/GameServerState.java
```diff
@@ -359,7 +362,7 @@ public void onCheatTriggered(CheatState.CheatType cheat, short playerId) {
                     break;
                 }
                 case WIN_LEVEL: {
-                    game.getGameController().endGame(playerId, true);
+                    stateManager.getState(PlayerState.class).endGame(true);
```
tonihele: Everything else is fine except this. Or debatable, but I think this should call the game controller. Everything should go through the official channels. And they should then trigger what they trigger, i.e. call clients that the game is won.

I do realize that this is inside the cheats that are not available in the multiplayer. But if we keep piling up game logic in both client and server, something is bound to be broken at some point. Not in sync.

All this is really a client - server architecture even if badly formed one.

What do you think?

ufdada: well to be honest it was a quick shot and more like a workaround to make the progress easier to test. I can change it back to the previous state. Would that be okay for you? the whole debriefing stuff needs to be fixed anyhow (also the normal winning condition didn't really work, thats why i used cheats to test).

i see there are also some conflicts i need to fix.

tonihele: Maybe it is better changed back so that we don't forget and continue a bad path. It is ok if we are missing some pieces, better than hacks

ufdada:
Oh boy, this was quite a merge, i hope i didn't miss something

tonihele:
> Oh boy, this was quite a merge, i hope i didn't miss something

Looked coherent still, is it ready?

ufdada:
> > Oh boy, this was quite a merge, i hope i didn't miss something
> 
> Looked coherent still, is it ready?

Looks good so far, compiles and runs. Found a small bug with reading and writing the level progress in the settings. But now it works.

For me it's ready

tonihele (approved):

ufdada:
I tried to fix the issues that deep source was complaining about, i hope this is okay for you

PR #526, opened 2026-08-19
Fix for wide screen in main menu

ufdada:
Since the original main menu has a fixed resolution and aspect ratio, its not noticable that the actual room is unfinished (has no model or texture in the back).
While our current implementation is exactly like the original, i still would implement a little fix so the "black hole" isn't noticble while (De-)Briefing or looking at the Extras-Menu.

It basically just clones the original asset and moves it back one tile. 
Since the asset was made to repeat, it's looks quite decent.
The frontend-Menu is anyway just used in the frontend, it should be okay having a "hack" like this.


tonihele (approved):

tonihele:
Thanks for your contribution!

PR #531, opened 2026-08-30
Modelviewer fixes

ufdada:
Some small fixes to make the modelviewer a little bit more usable.

the comment in build is just for easier starting the modelviewer.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,13 +26,18 @@ allprojects {
 application {
     applicationName = 'OpenKeeper'
     mainClass = 'toniarts.openkeeper.Main'
-    //mainClass = 'toniarts.openkeeper.tools.modelviewer.ModelViewer'
     if (System.getProperty("os.name").toLowerCase().contains("mac")) {
         applicationDefaultJvmArgs = ["-Dvisualvm.display.name=OpenKeeper", "-XstartOnFirstThread"]
     }
     else {
         applicationDefaultJvmArgs = ["-Dvisualvm.display.name=OpenKeeper"]
     }
+}
+
+task modelViewer(type: JavaExec) {
+    group = "application"
+    mainClass = "toniarts.openkeeper.tools.modelviewer.ModelViewer"
+    classpath = sourceSets.main.runtimeClasspath
 }
 
 configurations.configureEach {
--- a/build.gradle
+++ b/build.gradle
@@ -34,10 +34,10 @@ application {
     }
 }
 
-task modelViewer(type: JavaExec) {
-    group = "application"
-    mainClass = "toniarts.openkeeper.tools.modelviewer.ModelViewer"
+tasks.register("modelViewer", JavaExec) {
+    group = ApplicationPlugin.APPLICATION_GROUP
     classpath = sourceSets.main.runtimeClasspath
+    mainClass = 'toniarts.openkeeper.tools.modelviewer.ModelViewer'
 }
 
 configurations.configureEach {
--- a/build.gradle
+++ b/build.gradle
@@ -26,6 +26,7 @@ allprojects {
 application {
     applicationName = 'OpenKeeper'
     mainClass = 'toniarts.openkeeper.Main'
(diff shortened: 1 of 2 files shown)


src/main/java/toniarts/openkeeper/tools/modelviewer/ModelViewer.java
```diff
@@ -187,6 +188,14 @@ public static void main(String[] args) {
         }
 
         ModelViewer app = new ModelViewer();
+
+        AppSettings settings = new AppSettings(true);
```
tonihele, line 194: Heh yeah.. yep yep. We would need to add the jme-awt-dialogs (meaning that one would be asked about the graphic settings when this tool is started). But I suppose that is haram to the Android pursuits. And the main game doesn't need that. We should eventually separate this tool to a sub-project, then we can maybe have the dependency separately and leave it out from the regular build....

build.gradle
```diff
@@ -26,6 +26,7 @@ allprojects {
 application {
     applicationName = 'OpenKeeper'
     mainClass = 'toniarts.openkeeper.Main'
+    //mainClass = 'toniarts.openkeeper.tools.modelviewer.ModelViewer'
```
tonihele: Is it possible to have a separate Gradle task to achieve this? Commented out code in code i.e. build files isn't really a wanted feature. Now one can run it from IDEs quite easily, but we should have a separate Gradle task for the model viewer, right?

src/main/java/toniarts/openkeeper/tools/modelviewer/ModelViewer.java
```diff
@@ -448,7 +457,7 @@ public void onSelectionChanged(Object selection) {
             }
             case MAPS: {
                 // Load the selected map
-                KwdFile kwd = getKwdFile((String) selection);
+                KwdFile kwd = getKwdFile((String) selection + ".kwd");
```
tonihele: This might be broken now as I merged the Kwd changes?

ufdada: yes, already fixed

src/main/java/toniarts/openkeeper/tools/modelviewer/ModelViewer.java
ufdada: Well depends where you want to go with it. i just wanted to have a quick and easy fix. I didn't want to waste too much time into something most people don't see anyhow.

build.gradle
ufdada: i have to check. In maven something like this was quite easy

src/main/java/toniarts/openkeeper/tools/modelviewer/ModelViewer.java
tonihele: Yes, I'm just thinking out loud. Eventually. Not with this. I mean, I want to break up the whole project to few sub projects so it would hopefully be more clear... the dependencies and architecture...

build.gradle
tonihele: This at least worked:
```
tasks.register("modelViewer", JavaExec) {
    group = ApplicationPlugin.APPLICATION_GROUP
    classpath = sourceSets.main.runtimeClasspath
    mainClass = 'toniarts.openkeeper.tools.modelviewer.ModelViewer'
}
```

I took it from a 3 years old thread. Dunno if it is the current way. Then I got:
image

ufdada: Damn it, you beat me to it :)

it seams there is also a slightly different way to implement this:

```
task modelViewer(type: JavaExec) {
    group = "application"
    mainClass = "toniarts.openkeeper.tools.modelviewer.ModelViewer"
    classpath = sourceSets.main.runtimeClasspath
}
```
it works for me, can you check if it is recognised in your ide?

tonihele: image

ufdada: okay, changed. Is this from netbeans?

tonihele: IntelliJ IDEA

tonihele (approved):

PR #532, opened 2026-08-30
animate lava with sine wave

ufdada:
The title says it all :)

Animate the lava with a sine wave. Technically also possible for water (the original had a setting for it).

Tried to match it as closly to the original (in wave height and speed).


tonihele (approved):

tonihele:
Nice work, thanks! Eventually perhaps with shaders. But I suppose our shader work is also a bit incomplete. Even without all the Android stuff going on

PR #536, opened 2026-09-04
implement decay mechanism for campaign with highlight

### LINKED ISSUE #527: Alternative textures (decay and highlight) for campaign level selection
When progressing the story completed levels show a decayed version and later in the game some levels get highlighted.

* Decay is stored in the _D texture files
* Blue Highlighting is stored in the _E texture files
* Highlighted textures blink for about 8 seconds and then stay blue
* While blue highlightning exists for all levels, ist only used after compeleting level 5 and it's not used for every level
* decay and highlight are only showed when level selection is possible (campaignselect screen)

|  Level active | Levels highlighted |
| --- | --- |
| 6a, b | 7, 8|
| 7 | 7, 8, 10 |
| 8 | 8, 10, 12 |
| 9 | 10, 12 |
| 10 | 12 |
| 11a, b, c | 11a, b, c, 12, 16 |
| 12 | - |
| 13 | 13, 16 |
| 14 | - |
| 15a, b | - |
| 16 | - |
| 17 | 17, 18 |
| 18 | - |
| 19 | - |
| 20 | - |

ufdada:
Fixes #527

### DIFF
--- a/src/main/java/toniarts/openkeeper/view/map/construction/RoomConstructor.java
+++ b/src/main/java/toniarts/openkeeper/view/map/construction/RoomConstructor.java
@@ -28,7 +28,7 @@ import toniarts.openkeeper.view.map.WallSection;
  */
 public abstract class RoomConstructor {
 
-    protected static AssetManager assetManager = null;
+    protected final AssetManager assetManager;
     protected final RoomInstance roomInstance;
     protected final boolean[][] map;
     protected final Point start;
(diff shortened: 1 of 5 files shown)


src/main/java/toniarts/openkeeper/view/map/construction/RoomConstructor.java
```diff
@@ -28,7 +28,7 @@
  */
 public abstract class RoomConstructor {
 
-    protected final AssetManager assetManager;
+    protected static AssetManager assetManager = null;
```
tonihele: Hmm, I would be little bit hesitant to set these to static. I think it should remain instance variable.

ufdada: I hope you like my new approach

tonihele: Looks good! Thanks!

tonihele (approved):

Issue #534, opened 2026-09-02
Debriefing Screens


ufdada:
The current debriefing is just bare bones.

The original game had a Debriefing for every different Leveltype and Gametype.
Also lists a few statistics about the game played.

Also related to #528 and #292

Normal Campaign level:

<img width="1917" height="1151" alt="Image" src="https://github.com/user-attachments/assets/08955ef1-ef06-4bb3-b567-b477a4c95cc9" />

Statistics about:
* Main
  * Overall ranking for the level
  * Overall ranking
  * Special objects found
* Secondary
  * Level won
  * Used time
  * Enemy creatures killed
  * Heroes killed
  * Claimed rooms
  * Claimed land
  * Gold mined
  * Mana saved
  * Objects created
  * Creature amount
  * Converted creatures
  * Creature level trained
  * Level attempts

Special Level:

<img width="1898" height="1164" alt="Image" src="https://github.com/user-attachments/assets/d6d69507-8433-4e41-bbb5-cb1e8e9c6e52" />

Statistics about:

* Like normal campaign, but without overall ranking

MPD Level

<img width="2003" height="1137" alt="Image" src="https://github.com/user-attachments/assets/2352a2ff-9fac-4a35-939e-231c20954f1b" />

* Main 
  * Enemy creatures killed
  * Heroes killed
  * Claimed rooms
  * Claimed land
  * Gold mined
  * Mana saved

Skirmish / Multiplayer:

<img width="1877" height="1130" alt="Image" src="https://github.com/user-attachments/assets/d42a3611-734d-4361-b630-0c8a133d1c56" />

Like the normal level, but without Overall ranking and Level attempts and statistic for every player (4 max?)

PR #538, opened 2026-09-06
Debriefing fix

ufdada:
Shows the debrief screens for all game types

Features:
* Same camera position as before leaving the main menu into a game level (not just the starting position)
* Re-Enabling EndgameState (wasn't used before)
* Continue playing after Winning a level is possible (option for debriefing in the option menu!)

Also Fixed:
* Several typos
* campaign progress for levels with variation wasn't reset properly
* campaign can't be progressed, because the next level isn't set as playable

Not in scope:
* different debriefing screens for each game type
* Player statistics

Helps #534

### DIFF
--- a/assets/Interface/MainMenu.xml
+++ b/assets/Interface/MainMenu.xml
@@ -232,7 +232,7 @@
 
             <panel height="26px" />
             <!--Main objectives-->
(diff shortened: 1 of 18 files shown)


tonihele (approved):
Damn, now we have to implement creatures attacking objects/doors/etc to get them actually destroy your dungeon heart...

PR #539, opened 2026-09-09
Improve (De)Briefing screens

ufdada:
* Objective images are now larger (25% of screen)
* Title and Subtile have now a glow

grafik

Here a comparision to the original game:
grafik

### DIFF
--- a/src/main/java/toniarts/openkeeper/gui/nifty/effect/TextGlowEffect.java
+++ b/src/main/java/toniarts/openkeeper/gui/nifty/effect/TextGlowEffect.java
@@ -31,22 +31,51 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 /**
- * Renders a subtle glow around the element's text. The glow is drawn, with the
- * configured color and alpha, as concentric rings around the text so that it
- * appears as a soft halo behind the actual text. Use in conjunction with
- * {@code onActive}, {@code timeType="infinite"} and {@code neverStopRendering="true"}
- * to render the effect on every frame while the element's screen is active.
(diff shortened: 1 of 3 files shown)


src/main/java/toniarts/openkeeper/gui/nifty/effect/TextGlowEffect.java
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * Renders a subtle glow around the element's text. The glow approximates a
+ * Gaussian blur by drawing the text multiple times at all offsets within a
+ * small radius, with the alpha falling off like a normal distribution as the
+ * offset grows. This produces a soft, blurred halo behind the actual text. Use
+ * in conjunction with {@code onActive}, {@code timeType="infinite"} and
+ * {@code neverStopRendering="true"} to render the effect on every frame while
+ * the element's screen is active.
+ *
+ * @author Toni Helenius <helenius.toni@gmail.com>
```
tonihele, line 42: Thanks for giving me credit :D

tonihele (approved):
Cool, I think I tried to do the glow thingie a couple of times and never could get it done

src/main/java/toniarts/openkeeper/gui/nifty/effect/TextGlowEffect.java
ufdada: The AI thought it would fit best 😆

ufdada:
> Cool, I think I tried to do the glow thingie a couple of times and never could get it done

Even with the help of AI it took me quite some time. I think i had at least 12 iterations before i had something that actually worked

PR #540, opened 2026-09-12
general fixes

Trass3r:


assets/Shaders/Gui/ProgressIndicator.frag
```diff
@@ -1,15 +1,21 @@
-#import "Common/ShaderLib/GLSLCompat.glsllib"
+#extension GL_ARB_separate_shader_objects : enable
+#extension GL_ARB_explicit_uniform_location : enable
+
+precision highp float;
+precision highp sampler2D;
```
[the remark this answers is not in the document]
Trass3r: They are valid and simply ignored on desktop.

src/main/java/toniarts/openkeeper/game/state/MainMenuScreenController.java
```diff
@@ -141,6 +135,8 @@ public MainMenuScreenController(MainMenuState state, Nifty nifty) {
         this.state = state;
         this.nifty = nifty;
 
+        if (Main.isAudioDisabled())
```
tonihele, line 138: Yeah, Nifty uses jME's AudioRenderer. Can't we just have a kinda a proxy implementation that either uses NULL or the actual one depending on our settings? Otherwise we have billion of the brittle things everywhere?

tonihele (commented):
Otherwise ok but the audio handling scares me a bit. Meaning that we should handle it on some global level instead micro-manage every place we do some remotely audio related.

Nifty and jME share the audio renderer. In Nifty it seems it can't be changed on run time (well, maybe with reflection...). But we could just proxy the whole AudioRenderer, or rather implement our own that delegates to the real one if sounds are on. Or then just doesn't do anything if on sounds.

I didn't check if jME handle NULL renderer differently and this facade would actually hinder it... But I think we can assume that most of the time sounds are on... So any minor hiccups could be perhaps tolerated...

src/main/java/toniarts/openkeeper/game/state/MainMenuScreenController.java
Trass3r: Yeah it doesn't check for settings.audioRenderer==null. Nasty.

PR #541, opened 2026-09-14
Fog of war system

### LINKED ISSUE #139: fog of war
Have some idea how its realized in game?

ufdada:
Fixes #139 

Implements a fog of war system close to the original.

But due to it's complexity needs some good testing.

It is also a base for the minimap and lightning system

For anyone interested, here is the intial design concept:
[fog_of_war_design.md](https://github.com/user-attachments/files/32202820/fog_of_war_design.md)


ufdada:
Problems found so far:

- [x] Water and Lava in unexplored areas don't have their liquid texture
- [x] Objects aren't visible when explored
- [x] Rooms (Portal, herogate) are visible through FOW 
- [x] Outer perimeter always impenetrable

src/main/java/toniarts/openkeeper/view/PlayerMapViewState.java
```diff
@@ -84,10 +103,28 @@ public PlayerMapViewState(Main app, final IKwdFile kwdFile, final AssetManager a
             @Override
             protected void onLoadComplete() {
 
+                // Fog must be seeded before the initial geometry is built, since
+                // the renderer consults it while assembling the map
+                fogOfWarController.seedLevelStart();
```
tonihele, line 108: Shouldn't this be also inside the thread?

tonihele (approved):
I suppose it is alright, just not the way I visioned :D I guess the main difference is that it runs on the client, whereas I imagined that it would be server driven. Whether it matters really I suppose is multiplayer cheating and if we really go big on maps, I suppose also network performance. Nice work

ufdada:
> I suppose it is alright, just not the way I visioned :D I guess the main difference is that it runs on the client, whereas I imagined that it would be server driven. Whether it matters really I suppose is multiplayer cheating and if we really go big on maps, I suppose also network performance. Nice work

Thanks :)
It's not impossible to make this server driven ;). For now i wanted to keep it simple.

src/main/java/toniarts/openkeeper/view/PlayerMapViewState.java
ufdada: You mean in the maploaderthread? i fear i just would create a race (see PlayerMapViewState.update) without synchronizing the objects that get modified constantly. So i traded this potential race with it being single threaded.

Or do you have a better idea? I am a novice when it comes to multi-threading and AI isn't helping either 😉

tonihele: Yeah that I meant, but the code is kinda complicated so I can't really tell... The caller already seems to be in some thread...

PR #552, opened 2026-09-15
Reimplement effects for claiming and digging

ufdada:
Also now with the correct models crumbeling. Still lots of progress needs to be done


ufdada:
Instead of calling the effect for the gems from the mainmenu, it is now called from the room controller instead.

I think we need a redesign how to call these effects from rooms, tiles, creatures, traps, things and so on. But this isn't the goal for this one.

tonihele (commented):
Very nice! Didn't actually try to run but eager to have this. The map effects are somewhat what I thought they would be, that they are completely handled on the client and in-sync when it re-draws the map.

The room effects are currently a server side thing it seems. Is there a reason why these can exist purely on the client side?

ufdada:
> The room effects are currently a server side thing it seems. Is there a reason why these can exist purely on the client side?

My initial thought was that is should propagate it to the clients, since it's not only your room that gets a effect (especially when there is no fog of war). Might be completely wrong.

That's what i meant with the redesign, we should use an overall concept how we handle these effects. I don't want to go into the direction of implementing every effect by hand.

ufdada:
Also you can try it in the modelviewer, most of them work quite well there.

